PR #1, opened 2015-11-03
Add README with ideas for how the library will work

paulcsmith:


paulcsmith:
Merging this in so people get an idea of what's going on and can submit feedback if they'd like

PR #2, opened 2015-11-09
Add Mailer, Email, and MandrillAdapter

paulcsmith:
Still needs some work, but this is a good foundation to work off of

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ defmodule MyApp.Mailer do
 end
 
 defmodule MyApp.Emails do
-  use Bamboo.Email
+  import Bamboo.Email
 
   def welcome_email do
     mail(
@@ -56,9 +56,9 @@ defmodule MyApp.Emails do
 
   def welcome_email do
     base_email
-    # Bulk update the email
-    |> struct(bcc: "someone@bar.com", from: "other_person@foo.com")
-    |> to("foo@bar.com", ["John Smith": "john@foo.com"])
+    # A way to bulk update email attributes. Still figuring out
+    |> update_attrs(bcc: "someone@bar.com", from: "other_person@foo.com")
+    |> to("foo@bar.com", %{name: "John Smith", email:"john@foo.com"})
     |> cc(author) # You can set up a custom protocol that handles different types of structs.
     |> subject("Welcome!!!")
     |> tag("welcome-email") # Imported by Bamboo.MandrillEmails
@@ -79,8 +79,9 @@ end
 
 defimpl Bamboo.Formatter, for: User do
   # Used by `to`, `bcc`, `cc` and `from`
-  def extract_email(%User{email: email, name: name}) do
-    [name => email]
+  def extract_email(user) do
+    fullname = "#{user.first_name} #{user.last_name}"
+    %{name: fullname, email: email}
   end
 end
 ```
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -1,0 +1,51 @@
+defmodule Bamboo.Mailer do
+  defmacro __using__(opts) do
+    quote bind_quoted: [opts: opts] do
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -79,8 +79,9 @@ end
 
 defimpl Bamboo.Formatter, for: User do
   # Used by `to`, `bcc`, `cc` and `from`
-  def extract_email(%User{email: email, name: name}) do
-    [name => email]
+  def extract_email(user) do
```
paulcsmith: This isn't implemented yet. I'll add it later

test/lib/bamboo/mandrill_adapter_test.exs
```diff
@@ -0,0 +1,98 @@
...
+    post "/api/1.0/messages/send.json" do
+      conn |> send_resp(200, "SENT") |> send_to_parent
+    end
+
+    defp send_to_parent(conn) do
+      parent = Agent.get(__MODULE__, fn(set) -> HashDict.get(set, :parent) end)
+      send parent, {:fake_mandrill, conn}
+      conn
+    end
+  end
```
paulcsmith: I'm going to work on extracting this to it's own module in `test/support` next. Should make this a lot simpler and easier to read, also it would be reusable for other adapters in the future

README.md
```diff
@@ -56,9 +56,9 @@ defmodule MyApp.Emails do
 
   def welcome_email do
     base_email
-    # Bulk update the email
-    |> struct(bcc: "someone@bar.com", from: "other_person@foo.com")
-    |> to("foo@bar.com", ["John Smith": "john@foo.com"])
+    # A way to bulk update email attributes. Still figuring out
+    |> update_attrs(bcc: "someone@bar.com", from: "other_person@foo.com")
```
BlakeWilliams: Is there a better name that `attrs`? `meta` maybe?

lib/bamboo/mandrill_adapter.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Bamboo.MandrillAdapter do
+  @send_message_path "api/1.0/messages/send.json"
+
+  def deliver(email, config) do
+    api_key = Keyword.fetch!(config, :api_key)
+    params = email |> convert_to_mandrill_params(api_key) |> Poison.encode!
+    HTTPoison.post("http://localhost:4001/#{@send_message_path}", params, headers)
```
paulcsmith: I'll extract this URI so it pulls from the ENV, but I'll do that in a separate PR. Don't want to make this one too big

README.md
paulcsmith: It's not really metadata though. I would be ok with `update_email` even. This isn't implemented in code yet though because I'm not sure about it. It might even make sense to not have it and instead just recommend using `struct` to update it

lib/bamboo/mailer.ex
```diff
@@ -0,0 +1,54 @@
...
+      @adapter adapter
+      @config config
+
+      def deliver(email) do
+        Bamboo.Mailer.deliver(@adapter, email, @config)
+      end
+    end
+  end
+
+  def deliver(_, %{from: nil}, config) do
```
BlakeWilliams: Would this make sense as an adapter implemented error instead of on `mailer` itself?

paulcsmith: Yeah I think that's a good idea. Some adapters might have different rules :+1:

lib/bamboo/mandrill_adapter.ex
BlakeWilliams: Change the URL?

paulcsmith: Yeah maybe I should just do it now :S

README.md
BlakeWilliams: :+1:

BlakeWilliams:
Added some minor feedback, but overall looked good to me.

paulcsmith:
@BlakeWilliams Thanks for the feedback! I'll tackle that after lunch :)

PR #40, opened 2016-01-23
Empty recipients

paulcsmith:
I'm not sure about this, but I'm going to open it up and think about it a bit

### DIFF
--- a/lib/bamboo/adapters/mandrill_adapter.ex
+++ b/lib/bamboo/adapters/mandrill_adapter.ex
@@ -11,24 +11,22 @@ defmodule Bamboo.MandrillAdapter do
 
       Here is the response:
 
-      #{inspect response}
+      #{inspect response, limit: :infinity}
 
 
-      Here are the params we sent:
+      Here are the params that were sent:
 
-      #{inspect Poison.decode!(params)}
+      #{inspect Poison.decode!(params), limit: :infinity}
       """
       %ApiError{message: message}
     end
   end
 
   def deliver(email, config) do
-    api_key = get_key(config)
-    params = email |> convert_to_mandrill_params(api_key) |> Poison.encode!
-    case request!(@send_message_path, params) do
-      %{status_code: status} = response when status > 299 ->
-        raise(ApiError, %{params: params, response: response})
-      response -> response
+    if email.to == [] and email.cc == [] and email.bcc == [] do
+      {:no_recipients, email}
+    else
+      send_email(email, config)
     end
   end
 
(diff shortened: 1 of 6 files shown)


lib/bamboo/adapters/mandrill_adapter.ex
```diff
@@ -11,24 +11,22 @@ defmodule Bamboo.MandrillAdapter do
...
-    api_key = get_key(config)
-    params = email |> convert_to_mandrill_params(api_key) |> Poison.encode!
-    case request!(@send_message_path, params) do
-      %{status_code: status} = response when status > 299 ->
-        raise(ApiError, %{params: params, response: response})
-      response -> response
+    if email.to == [] and email.cc == [] and email.bcc == [] do
+      {:no_recipients, email}
+    else
+      send_email(email, config)
```
paulcsmith: Maybe this should be `{:ok, email}` so you can pattern match on the result...? Then I'd need to add `deliver!` and `deliver_async!` I think

lib/bamboo/mailer.ex
```diff
@@ -3,6 +3,26 @@ defmodule Bamboo.Mailer do
...
+
+      To - #{inspect email.to}
+      Cc - #{inspect email.cc}
+      Bcc - #{inspect email.bcc}
+
+      Full email - #{inspect email, limit: :infinity}
+      """
+      %NoRecipientError{message: message}
+    end
+  end
```
paulcsmith: Either way I think I want to keep this. to, cc and bcc shouldn't be set to nil.

lib/bamboo/email.ex
```diff
@@ -1,8 +1,8 @@
 defmodule Bamboo.Email do
   defstruct from: nil,
-      to: nil,
-      cc: nil,
-      bcc: nil,
+      to: [],
+      cc: [],
+      bcc: [],
```
paulcsmith: Actually I think this should go back to nil and it should raise it all recipients are nil. That would mean you forgot to set something

lib/bamboo/adapters/mandrill_adapter.ex
stevedomin: Being able to pattern match on the result seems useful. As you said you would need to add `deliver!` and `deliver_async!` but that's probably fine.

lib/bamboo/email.ex
paulcsmith: After thinking about this more, I think this should be how it's handled. If it starts as nil and you set an list then you probably meant to do that. I can see it going either way though

Here are a couple scenarios:

1) You are sending emails to people subcribed to a post. In your tests in dev you set up a post and have people subscribed and everything works just fine. Then in production you have posts with no subscribers. Would you want it to raise? I think I would expect it to just not send an email

2) You want to send to just one user, in which case you would not have a list. The only error I can imagine would be accidentally passing in nil. e.g. `Emails.successfully_published(article.author)` where the articles author is nil. In which case I *would* expect it to raise. 

So I think I'm going to go ahead with keep the default addresses as nil, and raising only if all of them are nil, and nooping if the the recipients are an empty array

paulcsmith:
Closed in favor of #67

PR #69, opened 2016-02-11
Relax httpoison dependency

sorentwo:
The dependency was locked to `~> 0.7.4`, which prevents the library from being used with any library or application using a version of httpoison that isn't `0.7.X`.

Specifically, I'm unable to use this in an application that also uses Hound, as there is a dependency conflict. Nothing seems to mandate the `0.7` series, and all tests pass after the change.


paulcsmith:
For some reason this is not passing on CI, but works fine locally. I'll have to take a look at this later. Might be because the FakeMandrillServer doesn't close the Cowboy server on_exit

paulcsmith:
@sorentwo Thanks for the PR!

PR #71, opened 2016-02-15
Deliver later strategies

### LINKED ISSUE #59: Don't use Task.async in deliver_async
`Task.async` is linking the task process with the parent process so if the latter dies first it will kill the former.

For example, with Phoenix it would mean that a HTTP request could finish before we actually had a chance to send the email.

Elixir 1.2 introduced [Task.async_no_link](http://elixir-lang.org/docs/stable/elixir/Task.Supervisor.html#async_nolink/2) which could be used here.

Happy to submit a PR if you think that's the way to go.

paulcsmith:
Closes #59 

@sorentwo @stevedomin You two were both interested in this I believe. What I did was move the deliver_later from the adapters to "deliver_later_strategies". So now you could create whatever strategy you want for deliver emails in the background. I think this simplifies adapters, and makes it easier to customize how you want to do background delivery. This also changes the default deliver_later strategy from using `Task.async` to using a `Task.Supervisor`.

LMK if you have any thoughts on this. @sorentwo With this new format you wouldn't need to create a new adapter, instead you would create a `GenServerStrategy` that you configure with `deliver_later_strategy: GenServerStrategy`.

### DIFF
--- a/lib/bamboo/adapter_behaviour.ex
+++ b/lib/bamboo/adapter_behaviour.ex
@@ -1,8 +1,6 @@
 defmodule Bamboo.Adapter do
-  @moduledoc """
+  @moduledoc ~S"""
   Use this behaviour when creating adapters to be used by Bamboo.
-
-  Accepts an email, and the config that was set for the mailer.
 
   ## Example
 
@@ -13,15 +11,20 @@ defmodule Bamboo.Adapter do
           deliver_the_email_somehow(email)
         end
 
-        def deliver_later(email, config) do
-          # You could also add the email to a GenServer or ExQ for delivery.
-          Task.async fn ->
-            Bamboo.CustomAdapter.deliver(email, config)
+        def handle_config(config) do
+          # Return the config if nothing special is required
+          config
+
+          # Or you could require certain config options
+          if Map.get(config, :smtp_username) do
+            config
+          else
+            raise "smpt_username is required in config, got #{inspect config}"
           end
         end
       end
   """
 
   @callback deliver(%Bamboo.Email{}, %{}) :: any
-  @callback deliver_later(%Bamboo.Email{}, %{}) :: Task.t
+  @callback handle_config(map) :: map
 end
--- a/lib/bamboo/adapters/local_adapter.ex
+++ b/lib/bamboo/adapters/local_adapter.ex
@@ -28,9 +28,22 @@ defmodule Bamboo.LocalAdapter do
     SentEmail.push(email)
   end
 
-  @doc "Adds email to Bamboo.SentEmail. Returns a Task that can be awaited on"
-  def deliver_later(email, _config) do
-    deliver(email, nil)
-    Task.async(fn -> :ok end)
+  def handle_config(config) do
+    case config[:deliver_later_strategy] do
+      nil ->
+        Map.put(config, :deliver_later_strategy, Bamboo.DeliverImmediatelyStrategy)
+      Bamboo.DeliverImmediatelyStrategy ->
+        config
+      _ ->
+        raise ArgumentError, """
+        Bamboo.LocalAdapter requires that the deliver_later_strategy is
+        Bamboo.DeliverImmediatelyStrategy
+
+        Instead it got: #{inspect config[:deliver_later_strategy]}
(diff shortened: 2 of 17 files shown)


lib/bamboo/adapter_behaviour.ex
```diff
@@ -13,15 +11,20 @@ defmodule Bamboo.Adapter do
...
+          else
+            raise "smpt_username is required in config, got #{inspect config}"
           end
         end
       end
   """
 
   @callback deliver(%Bamboo.Email{}, %{}) :: any
-  @callback deliver_later(%Bamboo.Email{}, %{}) :: Task.t
+  @callback handle_config(map) :: map
```
paulcsmith: Adapters can set default config options, or raise if things are missing

sorentwo:
@paulcsmith Sorry I'm a bit slow to reply here! I think the strategy of letting users implement their own async delivery is solid. For the record, after all of our discussion, I ended up with a mailer that always delivers async using `Task.Supervisor`. Here's the mailer:

```elixir
defmodule Px.Mailer do
  @supervisor Px.Mailer.Supervisor

  def deliver(email) do
    adapter = Map.fetch!(config, :adapter)
    parent  = self()

    Task.Supervisor.start_child(@supervisor, fn ->
      adapter.deliver(email, config, parent)
    end)
  end

  defp config do
    Application.get_env(:px, :mailer) |> Enum.into(%{})
  end
end
```

And the `Task.Supervisor` is started as a supervisor by the application. It's crazy how simple it is (when it isn't a library where you're trying to provide helpful error messages especially).

paulcsmith:
Very nice! I'm glad you got something up and running! You may run in to race conditions with the test adapter since it's running delivery in the background. The helpers use assert_received. I'm considering making them assert_receive with a short timeout though. That might help when people run in the background without using the DeliverImmediatelyStrategy

Let me know how your setup works out for you :)

lib/bamboo/adapters/local_adapter.ex
```diff
@@ -28,9 +28,22 @@ defmodule Bamboo.LocalAdapter do
...
-    deliver(email, nil)
-    Task.async(fn -> :ok end)
+  def handle_config(config) do
+    case config[:deliver_later_strategy] do
+      nil ->
+        Map.put(config, :deliver_later_strategy, Bamboo.DeliverImmediatelyStrategy)
+      Bamboo.DeliverImmediatelyStrategy ->
+        config
+      _ ->
+        raise ArgumentError, """
```
stevedomin: Why does that have to be the case? You should be able to use a deliver later strategy even with the local adapter, no?

paulcsmith: You’re right. It probably makes sense to allow any strategy. My one concern was that if someone used some kind of background job library, and something went wrong, or the jobs weren’t run, then the email wouldn’t show up in Bamboo.SentEmail. Now that I think about it though, that’s probably good. If you use some other kind of delivery strategy and it’s not working, you probably want to know about it.

So I think the LocalAdapter should allow any deliver_later_strategy, but what should the default be? Maybe it can be TaskSupervisorStrategy just like everything else. Then the adapter wouldn’t need anything special in handle_config at all.

stevedomin: Yep, I think it make sense to use `TaskSupervisorStrategy` by default

paulcsmith: :+1: I'll tackle that sometime this week, unless you want to open a PR :) If you don't have time though that's a-ok

lib/bamboo/strategies/deliver_immediately_strategy.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Bamboo.DeliverImmediatelyStrategy do
```
stevedomin: Minor point, what do you think of `ImmediateDeliveryStrategy` as a module name?

paulcsmith: I’m not sure why, but I like ImmediateDeliveryStrategy a lot more. That’s a great suggestion!

PR #80, opened 2016-02-17
Allow any strategy in local adapter

stevedomin:
From discussion in https://github.com/paulcsmith/bamboo/pull/71/files#r53177410, we want to allow any strategy in the local adapter.

Closes https://github.com/paulcsmith/bamboo/issues/79

### DIFF
--- a/lib/bamboo/adapters/local_adapter.ex
+++ b/lib/bamboo/adapters/local_adapter.ex
@@ -28,22 +28,5 @@ defmodule Bamboo.LocalAdapter do
     SentEmail.push(email)
   end
 
-  def handle_config(config) do
-    case config[:deliver_later_strategy] do
-      nil ->
-        Map.put(config, :deliver_later_strategy, Bamboo.DeliverImmediatelyStrategy)
-      Bamboo.DeliverImmediatelyStrategy ->
-        config
-      _ ->
-        raise ArgumentError, """
-        Bamboo.LocalAdapter requires that the deliver_later_strategy is
-        Bamboo.DeliverImmediatelyStrategy
-
-        Instead it got: #{inspect config[:deliver_later_strategy]}
-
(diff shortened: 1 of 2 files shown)


lib/bamboo/adapters/local_adapter.ex
```diff
@@ -32,18 +32,8 @@ defmodule Bamboo.LocalAdapter do
     case config[:deliver_later_strategy] do
       nil ->
         Map.put(config, :deliver_later_strategy, Bamboo.DeliverImmediatelyStrategy)
```
paulcsmith: I actually think we can just return the config as-is and use `TaskSupervisorStrategy` that is set by default in the mailer: https://github.com/paulcsmith/bamboo/blob/master/lib/bamboo/mailer.ex#L145

test/lib/bamboo/local_adapter_test.exs
```diff
@@ -19,15 +19,14 @@ defmodule Bamboo.LocalAdapterTest do
     assert SentEmail.all == [email]
   end
 
-  test "handle_config makes sure that the DeliverImmediatelyStrategy is used" do
+  test "use DeliverImmediatelyStrategy if strategy is unspecified" do
```
paulcsmith: And then we can remove this test completely since `handle_config` would just return the config as-is

paulcsmith:
@stevedomin Had just a couple comments

stevedomin:
Fixed.

What do you think of adding a default for `handle_config` to `AdapterBehaviour`?

paulcsmith:
@stevedomin I'm not against it, but I'm not sure how much value it provides. If the default is `def handle_config(config), do: config` then I don't think it's worth it at this point.

Also, thanks for the PR :) I'll merge it in as soon as CI passes

PR #86, opened 2016-02-21
[RFC] Reorganise codebase to be more consistent

stevedomin:
While working on a new adapter I found that the code organisation wasn't
necessarily consistent with what can be found in the Elixir codebase,
Ecto or Phoenix.

I opened the PR to kick start the discussion but there are still a couple of things I'm not happy with:
* The naming of the "Test" adapters leads to some weirdness: test file is named `test_test.exs` and the module name is TestTest. The purpose of the Local and Test adapter seems very similar so  I'm wondering if there isn't something we could do there. I've started playing with some ideas here.
* `Bamboo.SentEmail` is only useful for  `Bamboo.Adapters.Local` so the module should probably be moved closer to the adapter (e.g. `lib/adapters/local/sent_email.ex`)
* I'm also want to nest the strategies under `Bamboo.Strategies`. I'm actually thinking that maybe we should just get rid of the whole deliver/deliver_later and just have deliver but with a behaviour dependent on the strategy used. What do you think?

This is obviously a breaking change but since Bamboo hasn't reached 1.0 yet, now is the time :)

@paulcsmith would love to have your feedback on that

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Bamboo breaks email creation and email sending in to two separate modules.
 ```elixir
 # In your config/config.exs file
 config :my_app, MyApp.Mailer,
-  adapter: Bamboo.MandrillAdapter,
+  adapter: Bamboo.Adapters.Mandrill,
   api_key: "my_api_key"
 
 # Somewhere in your application
@@ -61,7 +61,7 @@ end
 defmodule MyApp.Emails do
   # Adds a `render` function for rending emails with a Phoenix view
   use Bamboo.Phoenix, view: MyApp.EmailView
-  import Bamboo.MandrillEmails
+  import Bamboo.Adapters.Mandrill
 
   def welcome_email do
     base_email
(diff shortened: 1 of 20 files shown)


stevedomin:
Re: strategies: another way to look at it (I think that's what you had in mind) is that strategies are only useful for later delivery, in which case we can simplify the whole thing (`Bamboo.DeliverLaterStrategyBehaviour` => `Bambo.Strategy`)  and make it consistent too. I'll push some code shortly.

stevedomin:
I've just pushed a new version with the changes around strategy, I think I like it better this way, it's probably useful to have `deliver` and `deliver_later` as different, rather than always specify the strategy you want to use.

paulcsmith:
@stevedomin Thanks for putting all the work in to this, but I don't think I'll merge this one in. I intentionally did not nest things in a `Strategies` or `Adapters` namespace for the same reason that Phoenix has `ThingController` and `ThingView

PR #98, opened 2016-03-25
Add preview email plug

paulcsmith:


lib/bamboo/plug/email_preview_plug.ex
```diff
@@ -0,0 +1,66 @@
...
+
+      # In your config
+      config :my_app, MyApp.Mailer,
+        adapter: Bamboo.LocalAdapter
+
+      # In your router
+      defmodule MyApp.Router do
+        use Phoenix.Router # or use Plug.Router if you're not using Phoenix
+
+        forward "/sent_emails", to: Bamboo.EmailPreviewPlug
```
paulcsmith: Remember to add part about only doing this if `Mix.env == :dev`

lib/bamboo/plug/index.html.eex
```diff
@@ -0,0 +1,173 @@
...
+
+        <section class="email-preview-bodies-container">
+          <h3 class="email-preview-body-label">HTML body</h3>
+          <p class="email-preview-body">
+            <%= @selected_email.html_body %>
+          </p>
+
+          <h3 class="email-preview-body-label">Text Body</h3>
+          <p class="email-preview-body">
+            <%= @selected_email.text_body %>
```
paulcsmith: For this we need to convert \n to <br> tags. I think that should be enough, at least to start with

PR #106, opened 2016-04-01
Add SendGrid adapter

jsteiner:


lib/bamboo/adapters/sendgrid_adapter.ex
```diff
@@ -0,0 +1,160 @@
...
+    * Here are the config options that were passed in:
+
+    #{inspect config}
+    """
+  end
+
+  defp headers(api_key) do
+    [
+      {"Content-Type", "application/x-www-form-urlencoded"},
+      {"User-Agent", "bamboo/#{Bamboo.version}"},
```
paulcsmith: I think it might be better to remove the stuff from `mix.exs` and instead set a module attribute in this module `@sendgrid_adapter_version`. I think that's more clear. What do you think?

test/lib/bamboo/adapters/sendgrid_adapter_test.exs
```diff
@@ -0,0 +1,132 @@
...
+    assert params["subject"] == email.subject
+    assert params["text"] == email.text_body
+    assert params["html"] == email.html_body
+    assert Enum.member?(headers, {"authorization", "Bearer #{@config[:api_key]}"})
+    assert Enum.member?(headers, {"user-agent", "bamboo/#{Bamboo.version}"})
+  end
+
+  test "deliver/2 correctly formats recipients" do
+    email = new_email(
+      to: [{"To", "to@bar.com"}],
```
paulcsmith: Might be good to add a recipient with a `nil` or `""` name to test the `list_empty?` function. What do you think?

lib/bamboo/adapters/sendgrid_adapter.ex
```diff
@@ -0,0 +1,160 @@
...
+
+      # In config/config.exs, or config.prod.exs, etc.
+      config :my_app, MyApp.Mailer,
+        adapter: Bamboo.SendgridAdapter,
+        api_key: "my_api_key"
+
+      # Define a Mailer. Maybe in lib/my_app/mailer.ex
+      defmodule MyApp.Mailer do
+        use Bamboo.Mailer, otp_app: :my_app
+      end
```
paulcsmith: I told @mtwilliams we'd credit him for first pass at the Sendgrid adapter Could we add a short note below this like "Thanks for @mtwilliams for the first pass at the Bamboo.SendGridAdapter", or "Credits to @mtwilliams for the first SendGridAdapter". Not sure I like those exact phrases, but something like that.

paulcsmith:
Just a few comments. Looks really good though. I'm excited to have more adapters in :D

paulcsmith:
Oh and can you also update the README to point to this adapter?

lib/bamboo/adapters/sendgrid_adapter.ex
jsteiner: Perhaps. It seems a little odd to put that in the docs. How's the bottom of the README?

jsteiner: For example: https://github.com/thoughtbot/bamboo/pull/106/commits/538b0df9280771ba16a9c8e4178231bdae43352c

paulcsmith:
LGTM!

PR #107, opened 2016-04-01
Handle tests with emails delivered from other processes

### LINKED ISSUE #105: Allow users to set the test process to send to when testing
Sometimes the process that you send an email from is not the test process. This happens for feature tests with hound, or if you deliver an email inside of another process (like inside of a Task).

1) Add option to set process name to send to `use Bamboo.Test, process_name: :invitation_controller_test`. This would tell Bamboo to send the messages to a process with that name. It would also name the test process with that name.
2) This would only work with `async: false`. Blow up if they try to set a process name with `async: true`. Let them know why, and tell the user, that if they're doing feature tests they could consider doing a controller test if they really want async tests. Related to #104 

I imagine it would look something like

```elixir
defmacro __using__(process_name: process_name) do
  quote do
    setup tags do
      if tags[:async] do
        raise "passing process_name cannot be done for async test. Set async: false"
      else
         # If this is set, use this as the process to send to instead of `self`
        Application.put_env(:bamboo, test_process_name: process_name)
        Process.register(self, process_name)
        import Bamboo.Test
      end
    end
  end
end

defmacro __using__(_opts) do
  quote do
    import Bamboo.Test
    Application.delete_env(:bamboo, :test_process_name)
  end
end
```

- [x] Set process name
- [x] Use process name for assertion if it is set
- [x] Raise if async: true

paulcsmith:
Closes #105

### DIFF
--- a/lib/bamboo/test.ex
+++ b/lib/bamboo/test.ex
@@ -1,10 +1,20 @@
 defmodule Bamboo.Test do
+  @timeout 100
+
   @moduledoc """
   Helpers for testing email delivery
 
   Use these helpers with Bamboo.TestAdapter to test email delivery. Typically
   you'll want to **unit test emails first**. Then in integration tests use
   helpers from this module to test whether that email was delivered.
+
+  ## Note on sending from other processes
+
+  If you are sending emails from another process (for example, from inside a
+  Task or GenServer) you may need to use the `process_name` option when using
+  `Bamboo.Test`. See the docs `__using__/1` for an example.
+
+  For most scenarios you will not need the `process_name` option.
 
   ## In your config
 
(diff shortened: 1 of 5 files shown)


lib/bamboo/test.ex
```diff
@@ -92,13 +130,13 @@ defmodule Bamboo.Test do
   def assert_delivered_email(%Bamboo.Email{} = email) do
     import ExUnit.Assertions
     email = Bamboo.Mailer.normalize_addresses(email)
-    assert_received {:delivered_email, ^email}
+    assert_receive {:delivered_email, ^email}, @timeout
```
paulcsmith: If you send from another process we need to wait a bit so that we don't get false failures. I could probably extract the timeout as an option at some point, but I think this is ok for now

lib/bamboo/test.ex
```diff
@@ -65,8 +67,44 @@ defmodule Bamboo.Test do
   @doc """
   Imports Bamboo.Test and Bamboo.Formatter.format_email_address/2
   """
+  defmacro __using__(process_name: process_name) do
+    quote do
+      setup tags do
+        if tags[:async] do
```
paulcsmith: I could not figure out how to test this, since it would raise before the tests ever ran (which is the point :P) So I'm not really sure how to test it. Might not be worth it?

paulcsmith:
@jsteiner Could you give this a try on your project sometime before I merge this in? Just to make sure it really does solve the problem with the feature tests

lib/bamboo/test.ex
paulcsmith: Maybe I can do it with this: https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/test/ex_unit/assertions_test.exs#L591

paulcsmith:
Note to self: Add moduledocs for the `process_name` option :)

BlakeWilliams:
LGTM, love the error message for async tests with `process_name`!

paulcsmith:
Thanks @BlakeWilliams

PR #108, opened 2016-04-01
Clarify delivery

paulcsmith:
If someone accidentally calls the Bamboo.Mailer directly, they get a
helpful message. This also allows us to add some docs to the deliver_now
and deliver_later functions.

Also bumps the version so I can update some errors in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,9 @@ See the [Bamboo.Email] and [Bamboo.Formatter docs] for more info and examples.
 
 ## Using Phoenix Views and Layouts
 
-Phoenix is not required to use Bamboo. However, if you do use Phoenix, you can use Phoenix views and layouts with Bamboo. See [Bamboo.Phoenix](https://hexdocs.pm/bamboo/Bamboo.Phoenix.html)
+Phoenix is not required to use Bamboo. However, if you do use Phoenix, you can
+use Phoenix views and layouts with Bamboo. See
+[Bamboo.Phoenix](https://hexdocs.pm/bamboo/Bamboo.Phoenix.html)
 
 ## Previewing Sent Emails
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -235,7 +240,7 @@ See documentation for [Bamboo.Test] for more examples, and remember to use
 Bamboo.TestAdapter.
 
 [Bamboo.Test]: https://hexdocs.pm/bamboo/Bamboo.Test.html
-  
+
```
paulcsmith: Wrapping lines and removing whitespace

README.md
```diff
@@ -197,11 +199,14 @@ See [Bamboo.EmailPreviewPlug](https://hexdocs.pm/bamboo/Bamboo.EmailPreviewPlug.
...
-Mandrill offers extra features on top of regular SMTP email like tagging, merge vars, and scheduling emails to send in the future. See [Bamboo.MandrillHelper](https://hexdocs.pm/bamboo/Bamboo.MandrillHelper.html).
+Mandrill offers extra features on top of regular SMTP email like tagging, merge
+vars, and scheduling emails to send in the future. See
+[Bamboo.MandrillHelper](https://hexdocs.pm/bamboo/Bamboo.MandrillHelper.html).
 
 ## Testing
 
-You can use the Bamboo.TestAdapter along with [Bamboo.Test] to make testing your emails a piece of cake.
+You can use the Bamboo.TestAdapter along with [Bamboo.Test] to make testing your
+emails a piece of cake.
```
drapergeek: Minor: what do you think about using `easy` here instead of `a piece of cake`. I don't think the reference makes any more distinction and I could imagine a lot of non-english natives not understanding that temr.

lib/bamboo/mailer.ex
```diff
@@ -74,6 +79,29 @@ defmodule Bamboo.Mailer do
     end
   end
 
+  @doc """
+  Deliver an email right away
+
+  Call your mailer with `deliver_now/1` to send an email right away. Call
+  `deliver_later/1` if you want to send in the background to speed things up.
+  """
+  def deliver_now(_email) do
```
drapergeek: :+1: very nice

README.md
paulcsmith: I like the general idea. I’m not a fan of easy though because it’s one of those words user so much that I feel it’s lost it’s meaning. Maybe there is something else that can be used?

* make testing straightforward
* make testing less error prone
* make testing less of a headache

?

drapergeek: I'm fine with the first one.

paulcsmith: :+1:

PR #109, opened 2016-04-01
Better test helpers

paulcsmith:
I want to improve the tests for these, but I also want to get this opened up so I can get feedback on wording for the failure messages.

The problem with using the default ExUnit.Assertions helpers is that they included messages that weren't emails and it wasn't always clear what the problem was. So I've implemented these helpers to give better messages to users when things go wrong. It should hopefully help people debug their tests more quickly and effectively

### DIFF
--- a/lib/bamboo/test.ex
+++ b/lib/bamboo/test.ex
@@ -1,6 +1,8 @@
 defmodule Bamboo.Test do
   @timeout 100
 
+  import ExUnit.Assertions
+
   @moduledoc """
   Helpers for testing email delivery
 
@@ -147,10 +149,10 @@ defmodule Bamboo.Test do
   @doc """
   Checks whether an email was delivered.
 
-  Must be used with the Bamboo.TestAdapter or this will never pass. If a
-  Bamboo.Email struct is passed in, it will check that all fields are matching.
-
-  You can also pass a keyword list and it will check just the fields you pass in.
+  Must be used with the `Bamboo.TestAdapter` or this will never pass. In case you
+  are delivering from another process, the assertion waits up to 100ms before
+  failing. Typically if an email is successfully delivered the assertion will
+  pass instantly, so test suites will remain fast.
 
   ## Examples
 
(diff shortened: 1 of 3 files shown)


test/lib/bamboo/test_adapter_test.exs
```diff
@@ -44,20 +44,107 @@ defmodule Bamboo.TestAdapterTest do
...
+  test "assert_delivered_email with no delivered emails" do
+    sent_email = new_email(from: "foo@bar.com", to: ["foo@bar.com"])
+
+    try do
+      assert_delivered_email %{sent_email | to: "oops"}
+    rescue
+      error in [ExUnit.AssertionError] ->
+        assert error.message =~ "0 emails delivered"
+    else
+      _ -> flunk "assert_delivered_email should failed"
```
paulcsmith: This is necessary so that if the assertion didn't fail we raise. Otherwise it looks like the test passed.

test/lib/bamboo/test_adapter_test.exs
```diff
@@ -44,20 +44,107 @@ defmodule Bamboo.TestAdapterTest do
...
+    end
+  end
+
+  test "assert_delivered_email with no delivered emails" do
+    sent_email = new_email(from: "foo@bar.com", to: ["foo@bar.com"])
+
+    try do
+      assert_delivered_email %{sent_email | to: "oops"}
+    rescue
+      error in [ExUnit.AssertionError] ->
```
paulcsmith: This is some funky stuff. I got it from https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/test/ex_unit/assertions_test.exs#L591

It's how they test their assertion helpers in ExUnit

lib/bamboo/test.ex
```diff
@@ -90,23 +94,91 @@ defmodule Bamboo.Test do
...
+      email_message
+    end
+  end
+
+  defp delivered_emails_as_list do
+    delivered_emails |> add_asterisk |> Enum.join("\n")
+  end
+
+  defp add_asterisk(emails) do
+    for email <- emails do
```
BlakeWilliams: `emails |> Enum.map(" * #{inspect &1}")` maybe?

BlakeWilliams:
This looked good to me.

lib/bamboo/test.ex
```diff
@@ -90,23 +94,91 @@ defmodule Bamboo.Test do
...
   end
 
   @doc """
   Checks that no emails were sent.
   """
+  def assert_no_emails_delivered do
+    receive do
+      {:delivered_email, email} -> flunk_with_unexpected_email(email)
+    after
+      0 -> true
```
paulcsmith: Before I merge this in I want to wait for #105 to close. Then I can raise with a message that tells people to set a `refute_delivery_timeout` that is required when using `process_name` with Bamboo.Test. That way for most tests it has 0 timeout, but for process_name people can choose how long they wait

lib/bamboo/test.ex
paulcsmith: Yeah I like that!

lib/bamboo/test.ex
```diff
@@ -90,23 +94,91 @@ defmodule Bamboo.Test do
...
+      flunk """
+      There were 0 emails delivered to this process.
+
+      If you expected an email to be sent, try these ideas:
+
+        1) Make sure you call deliver_now/1 or deliver_later/1 to deliver the email
+        2) Make sure you are using the Bamboo.TestAdapter
+        3) Use the process name feature of Bamboo.Test. This will allow Bamboo.Test
+           to work across processes: use Bamboo.Test, process_name: :my_test_name
+        4) If you are writing an acceptance test through a headless browser, see
```
jsteiner: Should this matter with `https://github.com/thoughtbot/bamboo/pull/107`?

lib/bamboo/test.ex
```diff
@@ -90,23 +94,91 @@ defmodule Bamboo.Test do
...
+      flunk """
+      There were no matching emails.
+
+      No emails matched:
+
+        #{inspect email}
+
+      Delivered emails:
+
+      #{delivered_emails_as_list}
```
jsteiner: Should this be indented too?

paulcsmith: The problem was that if I indented in this string it would indent the first email, but the rest wouldn’t be indented. So I had to do the indenting in the delivered_emails_as_list function. Open to other ideas if you can think of something.

lib/bamboo/test.ex
paulcsmith: It does because if you are running acceptance test through a headless browser this is telling you to either write a controller test or use the process_name feature. 

I would like to guide people toward doing mostly controller tests and treating acceptance tests as super high level. So controller tests would check for sent emails, maybe check that some other stuff was done or saved to the db and acceptance tests check that it didn't fail and had something on the page.

I guess this is more of a personal preference, but I think this will lead to a faster/more maintainable test suite

paulcsmith: Maybe I can clarify this suggestion though? Any ideas on how to do that?

jsteiner: I'm cool with that preference, although it is odd to include it in this list of potential reasons it didn't work if there is no _technical_ reason for it.

paulcsmith: Yeah it's kind of iffy, because you can't use the async tests with them. So I'm trying to let people know that. Maybe I could rephrase it a bit?

```
If this is an acceptance test that is run through a headless browser, you can instead write a controller test, or set the process_name option as described in option 3.
```

I feel like that gives people the 2 options. They can either switch to a different type of test or set the process name option. So it doesn't really say which way is best, but does give you ways to handle it. What do you think?

lib/bamboo/test.ex
```diff
@@ -1,6 +1,8 @@
 defmodule Bamboo.Test do
   @timeout 100
```
paulcsmith: I think this should be configurable. Also should be renamed to `assert_timeout` since we have a separate timeout value for refute

paulcsmith: Changed my mind. Hardcoded at 100ms is fine for now (for assertion). I did add required configuration when refuting and using process_name though.

PR #118, opened 2016-04-12
Add Sparkpost adapter

andrewtimberlake:
I’ve written a basic Sparkpost adapter

### DIFF
--- a/lib/bamboo/adapters/sparkpost_adapter.ex
+++ b/lib/bamboo/adapters/sparkpost_adapter.ex
@@ -1,0 +1,162 @@
+defmodule Bamboo.SparkpostAdapter do
+  @moduledoc """
+  Sends email using Sparkpost's JSON API.
+
+  Use this adapter to send emails through Sparkpost's API. Requires that an API
+  key is set in the config. See `Bamboo.SparkpostHelper` for extra functions that
+  can be used by `Bamboo.SparkpostAdapter` (tagging, merge vars, etc.)
+
+  ## Example config
+
+      # In config/config.exs, or config/prod.exs, etc.
+      config :my_app, MyApp.Mailer,
+        adapter: Bamboo.SparkpostAdapter,
+        api_key: "my_api_key"
+
+      # Define a Mailer. Maybe in lib/my_app/mailer.ex
+      defmodule MyApp.Mailer do
+        use Bamboo.Mailer, otp_app: :my_app
+      end
+  """
+
+  @default_base_uri "https://api.sparkpost.com/"
+  @send_message_path "api/v1/transmissions"
+  @behaviour Bamboo.Adapter
+
+  defmodule ApiError do
+    defexception [:message]
+
+    def exception(%{params: params, response: response}) do
+      filtered_params = params |> Poison.decode! |> Map.put("key", "[FILTERED]")
+
+      message = """
+      There was a problem sending the email through the Sparkpost API.
+
+      Here is the response:
+
+      #{inspect response, limit: :infinity}
+
+
+      Here are the params we sent:
+
+      #{inspect filtered_params, limit: :infinity}
+      """
+      %ApiError{message: message}
+    end
+  end
+
+  def deliver(email, config) do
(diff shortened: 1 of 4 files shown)


paulcsmith:
@andrewtimberlake thank you so much! Adapters are so important to allowing people to use Bamboo. 

I'll try to review this this week, but before I merge it in can you let me know how much testing you've done in production? If none or not much it might be best to keep this in a separate repo and then I can link to it in the readme along with a note that says it's still in beta. What do you think?

andrewtimberlake:
I'm using it in production on a small app. Most of the code is adapted directly from the Mandril adapter with the changes driven by tests. I changed the API endpoint and the JSON payload.
Either way, I’m happy.

paulcsmith:
I've been thinking about this the past couple days and I think it would be best if you extracted this to a repo you own and then publish to hex. I would of course, LOVE to link to it in the readme. Here's my reasoning about why

* If you need to make changes you can make them right away
* I've been very busy lately so I think taking on another adapter that I don't use is a bit too time consuming for me right now. 
* Since my colleagues and I aren't using this adapter it's hard for us to guarantee that it works. We want to make sure that Bamboo works with the adapters it ships with and we can't do that easily with this one. Even though the adapters are unit tested, I try out bamboo on real projects before I publish to hex. 
* I'm considering moving the adapters to their own repos anyway, so it probably doesn't make sense to add more to Bamboo core. 

What do you think?

If this sounds good to you then I think the next steps are:

* Create repo for bamboo-sparkpost (or whatever you want to name it)
* Publish to hex
* Open an issue here in Bamboo with a link to the repo 
* I'll add a link to the readme :)

lib/bamboo/adapters/sparkpost_adapter.ex
```diff
@@ -0,0 +1,162 @@
+defmodule Bamboo.SparkpostAdapter do
+  @moduledoc """
+  Sends email using Sparkpost's JSON API.
+
+  Use this adapter to send emails through Sparkpost's API. Requires that an API
+  key is set in the config. See `Bamboo.SparkpostHelper` for extra functions that
+  can be used by `Bamboo.SparkpostAdapter` (tagging, merge vars, etc.)
```
paulcsmith: It looks like there is no SparkpostHeloer yet so maybe this could be removed

test/lib/bamboo/adapters/sparkpost_adapter_test.exs
```diff
@@ -0,0 +1,135 @@
...
+      pass: ["*/*"],
+      json_decoder: Poison
+    plug :match
+    plug :dispatch
+
+    def start_server(parent) do
+      Agent.start_link(fn -> HashDict.new end, name: __MODULE__)
+      Agent.update(__MODULE__, &HashDict.put(&1, :parent, parent))
+      Application.put_env(:bamboo, :mandrill_base_uri, "http://localhost:4001")
+      Plug.Adapters.Cowboy.http __MODULE__, [], port: 4001, ref: __MODULE__
```
paulcsmith: You may want to use the changes from this commit: https://github.com/thoughtbot/bamboo/commit/1de0bc49e624beef9a5607d139aafeb99489615e 

Previously there were intermittent failures because a port wouldn't fully close before a new test started. The changes in that commit ensure a free port is always used :)

paulcsmith:
I may actually pull this in and then extract all the adapters at a later point. I'll keep you posted. If you do create your own repo though, I'd be happy to link to it instead

lib/bamboo/adapters/sparkpost_adapter.ex
andrewtimberlake: I have added the helper

andrewtimberlake:
@paulcsmith I have extracted the Sparkpost Adapter into it’s own repo at https://github.com/andrewtimberlake/bamboo_sparkpost

paulcsmith:
Closed by #128 

Thanks a ton @andrewtimberlake :D

Issue #91, opened 2016-03-22
Add Bamboo.Phoenix.put_layout that accepts an atom for the template name


paulcsmith:
And use it for both the html and text layouts so you don't need to call `put_text_layout` and `put_html_layout`

paulcsmith:
@anellis This is a smaller one that could be useful. Right now we have functions for `put_text_layout` and `put_html_layout`. I'd also love to have `put_layout(emails, {MyApp.LayoutView, :email.html})`

This would basically call

```elixir
email
|> Bamboo.Phoenix.put_text_layout({layout, to_string(template) <> ".text")
|> Bamboo.Phoenix.put_html_layout({layout, to_string(template) <> ".html")
```

Might be a good started task if you're interested!

anellis:
@paulcsmith Awesome! I'll take a look and let you know if I have any questions

PR #122, opened 2016-04-15
Add a put_layout/2 function for text and html

anellis:
References #91

### DIFF
--- a/lib/bamboo/phoenix.ex
+++ b/lib/bamboo/phoenix.ex
@@ -14,9 +14,11 @@ defmodule Bamboo.Phoenix do
 
         def text_and_html_email_with_layout do
           new_email()
-          # You could also only set a layout for just html, or just text
-          |> put_html_layout({MyApp.LayoutView, "email.html"})
-          |> put_text_layout({MyApp.LayoutView, "email.text"})
+          # You could set just a text layout or just an html layout
+          |> put_text_layout
+          |> put_html_layout
+          # Or you can set a layout for both html and text at the same time
+          |> put_layout
           # Pass an atom to render html AND plain text templates
           |> render(:text_and_html_email)
         end
@@ -98,9 +100,16 @@ defmodule Bamboo.Phoenix do
     raise "function implemented for documentation only, please call: use Bamboo.Phoenix"
   end
 
-
   @doc """
   Sets the layout when rendering HTML templates
+
+  ## Example
+
+      def html_email_layout do
+        new_email
+        # Will use MyApp.LayoutView with email.html layout when rendering html emails
+        |> put_html_layout({MyApp.LayoutView, "email.html"})
+      end
   """
   def put_html_layout(email, layout) do
     email |> put_private(:html_layout, layout)
@@ -108,9 +117,35 @@ defmodule Bamboo.Phoenix do
 
   @doc """
   Sets the layout when rendering plain text templates
+
+  ## Example
+
+      def text_email_layout do
+        new_email
+        # Will use MyApp.LayoutView with email.text layout when rendering text emails
+        |> put_text_layout({MyApp.LayoutView, "email.text"})
+      end
   """
   def put_text_layout(email, layout) do
     email |> put_private(:text_layout, layout)
+  end
+
+  @doc """
+  Sets the layout for rendering plain text and HTML templates
+
+  ## Example
+
+      def text_and_html_email_layout do
+        new_email
+        # Will use MyApp.LayoutView with the email.html template for html emails
+        # Will use MyApp.LayoutView with the text.html template for text emails
+        |> put_layout({MyApp.LayoutView, :email})
+      end
+  """
+  def put_layout(email, {layout, template}) do
+    email
+    |> put_text_layout({layout, to_string(template) <> ".text"})
+    |> put_html_layout({layout, to_string(template) <> ".html"})
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


anellis:
@paulcsmith I noticed for `render/2` there's this:

```
Pass an atom as the template name to render HTML and plain text emails, e.g. :welcome_email. 
Use a string if you only want to render one type, e.g. "welcome_email.text" or "welcome_email.html"
```
Should we make the same option available for `put_layout`? So that you can do:
`put_layout(email, {Layout, "app.html"})` and it would do `put_html_layout` for you? Or do you think that's unnecessary since `put_html_layout/2` and `put_text_layout/2` are already public functions?

paulcsmith:
@anellis This is a great point. I've thought about this, but I think that what you did is good. It's probably good not to break the public API and I think using just one `put_layout` function a tab bit less clear. For example

```elixir
email
|> put_layout({LayoutView, "email.html"})
|> render(:the_email)
```

Would Bamboo use the email.html layout for both of them or just for the HTML version? You could assume that it uses it for the HTML version only, but I think it's a little unclear. So I think this PR is great as-is.

If anything I may rethink rendering so there is `render_text` `render_html` and `render`, but I'm not sure about that quite yet. That might be something for a later day :)

lib/bamboo/phoenix.ex
```diff
@@ -114,6 +112,15 @@ defmodule Bamboo.Phoenix do
   end
 
   @doc """
+  Sets the layout for rendering plain text and HTML templates
```
paulcsmith: While we have this open, could you add examples to the layout functions? Something simple like 

```elixir
# Will use MyApp.LayoutView with email.html and email.text layouts when rendering
put_layout(email, {MyApp.LayoutView, :email})
```

What do you think?

lib/bamboo/phoenix.ex
```diff
@@ -114,6 +112,15 @@ defmodule Bamboo.Phoenix do
...
+  """
+  def put_layout(email, {layout, template}) do
+    email
+    |> Bamboo.Phoenix.put_text_layout({layout, to_string(template) <> ".text"})
```
paulcsmith: I think you can call this as `|> put_text_layout` since it's in the same module

paulcsmith:
Just a couple small comments! Thanks for tackling this :)

anellis:
@paulcsmith That makes sense! I can see how having just the one put_layout function could be pretty confusing. 

I've updated the docs with examples and fixed that unnecessary reference to the current module.

lib/bamboo/phoenix.ex
```diff
@@ -98,22 +97,57 @@ defmodule Bamboo.Phoenix do
...
-
   @doc """
   Sets the layout when rendering HTML templates
+
+  ## Example
+
+      def html_email_layout do
+        new_email
+        |> from("myapp@example.com")
+        # Will use MyApp.LayoutView with email.html layout when rendering
```
paulcsmith: For `text` and `html` layout functions, what do you think of doing `...email.html layout when rendering html emails

lib/bamboo/phoenix.ex
```diff
@@ -98,22 +97,57 @@ defmodule Bamboo.Phoenix do
...
 
   @doc """
   Sets the layout when rendering plain text templates
+
+  ## Example
+
+      def text_email_layout do
+        new_email
+        |> from("myapp@example.com")
+        # Will use MyApp.LayoutView with email.text layout when rendering
```
paulcsmith: And `...email.text layout when rendering text emails`

lib/bamboo/phoenix.ex
```diff
@@ -15,8 +15,7 @@ defmodule Bamboo.Phoenix do
         def text_and_html_email_with_layout do
           new_email()
           # You could also only set a layout for just html, or just text
-          |> put_html_layout({MyApp.LayoutView, "email.html"})
-          |> put_text_layout({MyApp.LayoutView, "email.text"})
+          |> put_layout({MyApp.LayoutView, :email})
```
paulcsmith: How about adding separate things here for the different possibilities

```elixir
# You could set just a text layout or just an html layout
|> put_text_layout...
|> put_html_layout...
# Or you can set a layout for both html and text at the same time
|> put_layout...
```

lib/bamboo/phoenix.ex
```diff
@@ -98,22 +97,57 @@ defmodule Bamboo.Phoenix do
...
   end
 
   @doc """
+  Sets the layout for rendering plain text and HTML templates
+
+  ## Example
+
+      def text_and_html_email_layout do
+        new_email
+        |> from("myapp@example.com")
```
paulcsmith: We could probably remove the `from` in these examples. Even though a from is required to send an email I think it's ok to leave it out for the example

lib/bamboo/phoenix.ex
```diff
@@ -98,22 +97,57 @@ defmodule Bamboo.Phoenix do
...
+        # Will use MyApp.LayoutView with email.html and email.text layouts when rendering
```
paulcsmith: After look at it a bit more I think we could clarify it a bit

```
# Will use MyApp.LayoutView with the email.html template for html emails
# Will use MyApp.LayoutView with the text.html template for text emails
```

Is that too verbose, or do you think that adds more clarity?

paulcsmith:
Left a couple small comments about documentation. Other than that, I think it looks good to merge!

lib/bamboo/phoenix.ex
anellis: I think in documentation it's always better too be too verbose than to not provide enough info! I'll add that in

PR #125, opened 2016-04-17
Add Mailgun adapter

princemaple:
copied and modified from the sendgrid adapter

### DIFF
--- a/lib/bamboo/adapters/mailgun_adapter.ex
+++ b/lib/bamboo/adapters/mailgun_adapter.ex
@@ -77,8 +77,12 @@ defmodule Bamboo.MailgunAdapter do
   defp headers(config) do
     [
       {"Content-Type", "application/x-www-form-urlencoded"},
-      {"Authorization", "Basic #{Base.encode64(config.api_key)}"},
+      {"Authorization", "Basic #{auth_token(config)}"},
     ]
+  end
+
+  defp auth_token(config) do
+    Base.encode64("api:" <> config.api_key)
   end
 
   @mailgun_message_fields ~w(from to cc bcc subject text html)a
--- a/lib/bamboo/adapters/mailgun_adapter.ex
+++ b/lib/bamboo/adapters/mailgun_adapter.ex
@@ -81,7 +81,7 @@ defmodule Bamboo.MailgunAdapter do
     ]
   end
 
-  @mailgun_message_fields [:from, :to, :cc, :bcc, :subject, :text, :html]
+  @mailgun_message_fields ~w(from to cc bcc subject text html)a
 
   defp to_mailgun_body(%Email{} = email) do
     email
--- a/lib/bamboo/adapters/mailgun_adapter.ex
+++ b/lib/bamboo/adapters/mailgun_adapter.ex
@@ -1,0 +1,121 @@
+defmodule Bamboo.MailgunAdapter do
+  @moduledoc """
+  Sends email using Mailgun's API.
+
+  Use this adapter to send emails through Mailgun's API. Requires that an API
+  key and a domain are set in the config.
+
+  ## Example config
+
+      # In config/config.exs, or config.prod.exs, etc.
+      config :my_app, MyApp.Mailer,
+        adapter: Bamboo.MailgunAdapter,
+        api_key: "my_api_key",
+        domain: "your.domain"
+
+      # Define a Mailer. Maybe in lib/my_app/mailer.ex
+      defmodule MyApp.Mailer do
+        use Bamboo.Mailer, otp_app: :my_app
+      end
+  """
+
+  @base_uri "https://api.mailgun.net/v3/"
+  @behaviour Bamboo.Adapter
+
+  alias Bamboo.Email
+
+  defmodule ApiError do
+    defexception [:message]
+
+    def exception(%{params: params, response: response}) do
+      message = """
+      There was a problem sending the email through the Mailgun API.
+
+      Here is the response:
+
+      #{inspect response, limit: :infinity}
+
+
+      Here are the params we sent:
+
+      #{inspect params, limit: :infinity}
+      """
+      %ApiError{message: message}
+    end
+  end
+
+  def deliver(email, config) do
+    body = email |> to_mailgun_body |> Plug.Conn.Query.encode
+
+    case HTTPoison.post!(full_uri(config), body, headers(config)) do
+      %{status_code: status} = response when status > 299 ->
+        raise(ApiError, %{params: body, response: response})
+      response -> response
+    end
+  end
+
+  @doc false
+  def handle_config(config) do
+    for setting <- [:api_key, :domain] do
+      if config[setting] in [nil, ""] do
+        raise_missing_setting_error(config, setting)
+      end
+    end
+    config
+  end
+
+  defp raise_missing_setting_error(config, setting) do
+    raise ArgumentError, """
+    There was no #{setting} set for the Mailgun adapter.
+
+    * Here are the config options that were passed in:
+
+    #{inspect config}
+    """
+  end
+
+  defp headers(config) do
+    [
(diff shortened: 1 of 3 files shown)


princemaple:
screenshot 2016-04-18

tested working.

paulcsmith:
@princemaple Thank you so much. I started work on this, but obviously didn't finish yet. I'll be happy to review this and merge it in sometime this week!

princemaple:
@paulcsmith rebased this one too.

lib/bamboo/adapters/mailgun_adapter.ex
```diff
@@ -0,0 +1,125 @@
...
+  @base_uri "https://api.mailgun.net/v3/"
+  @behaviour Bamboo.Adapter
+
+  alias Bamboo.Email
+
+  defmodule ApiError do
+    defexception [:message]
+
+    def exception(%{params: params, response: response}) do
+      message = """
```
paulcsmith: Can you take a look at https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/mandrill_adapter.ex#L30 and the corresponding test https://github.com/thoughtbot/bamboo/blob/master/test/lib/bamboo/adapters/mandrill_adapter_test.exs#L140

I think we should do the same for this adapter so the API key isn't in the logs

lib/bamboo/adapters/mailgun_adapter.ex
```diff
@@ -0,0 +1,125 @@
...
+  end
+
+  defp raise_missing_setting_error(config, setting) do
+    raise ArgumentError, """
+    There was no #{setting} set for the Mailgun adapter.
+
+    * Here are the config options that were passed in:
+
+    #{inspect config}
+    """
```
paulcsmith: I like how you were able to use this error for both the api_key and domain! 👍

lib/bamboo/adapters/mailgun_adapter.ex
```diff
@@ -0,0 +1,125 @@
...
+  @mailgun_message_fields ~w(from to cc bcc subject text html)a
+
+  defp to_mailgun_body(%Email{} = email) do
+    email
+    |> Map.from_struct
+    |> combine_name_and_email
+    |> put_html_body(email)
+    |> put_text_body(email)
+    |> Map.take(@mailgun_message_fields)
+    |> Enum.filter(fn {_k, v} -> not v in [nil, "", []] end)
```
paulcsmith: What do you think about extracting a private function for this. Something like:

```elixir
|> remove_empty_fields
```

Thoughts?

lib/bamboo/adapters/mailgun_adapter.ex
```diff
@@ -0,0 +1,125 @@
...
+  defp combine_name_and_email(tuple) when is_tuple(tuple) do
+    case tuple do
+      {nil, email} -> email
+      {name, email} -> "#{name} <#{email}>"
+    end
+  end
+
+  defp put_html_body(body, %Email{html_body: html_body}), do: put_in(body[:html], html_body)
+
+  defp put_text_body(body, %Email{text_body: text_body}), do: put_in(body[:text], text_body)
```
paulcsmith: I would use `body.text` and `body.html` in most of these places. That way it raises if for some crazy reason those fields aren't set.

lib/bamboo/adapters/mailgun_adapter.ex
```diff
@@ -0,0 +1,125 @@
...
+
+  defp auth_token(config) do
+    Base.encode64("api:" <> config.api_key)
+  end
+
+  @mailgun_message_fields ~w(from to cc bcc subject text html)a
+
+  defp to_mailgun_body(%Email{} = email) do
+    email
+    |> Map.from_struct
```
paulcsmith: You *might* be able to remove this since the `Map.take` on line 96 will make it into a regular map

test/lib/bamboo/adapters/mailgun_adapter_test.exs
```diff
@@ -0,0 +1,132 @@
...
+    assert_receive {:fake_mailgun, %{params: params, req_headers: headers}}
+
+    assert params["from"] == elem(email.from, 1)
+    assert params["subject"] == email.subject
+    assert params["text"] == email.text_body
+    assert params["html"] == email.html_body
+
+    hashed_token  = Base.encode64("api:" <> @config.api_key)
+
+    assert Enum.member?(headers, {"authorization", "Basic #{hashed_token}"})
```
paulcsmith: What do you think of `assert {"authorization", "Basic #{hashed_token}"} in headers` ? I think that's a bit more clear

paulcsmith:
Nice job and thank you so much! Left a couple small comments, but overall this looks super awesome!

lib/bamboo/adapters/mailgun_adapter.ex
princemaple: Ha! I used `body[...]` here because they are precisely not there yet when this call happens. And I thought about moving the `Map.take` call upwards but `Map.take` doesn't create the key in the new map if it doesn't exist in the map being taken from.

paulcsmith: Oh I see! That makes sense now. Since they're not deeply nested keys, maybe Map.put would be more clear?

lib/bamboo/adapters/mailgun_adapter.ex
princemaple: I actually need it to be a map before calling `combine_name_and_email`.

This line

```elixir
Enum.reduce([:from, :to, :cc, :bcc], map, fn key, acc ->
  Map.put(acc, key, combine_name_and_email(map[key])) # <---- this line
end)
```

requires the `map` to implement `Access` protocol, which structs don't.

paulcsmith: Ah, that makes sense!

lib/bamboo/adapters/mailgun_adapter.ex
princemaple: @paulcsmith Mailgun authentication token is actually in the request headers.
This error message logs the params that are sent to the endpoint plus the response, hence the auth header is not disclosed.

below is an intentionally failed test case. you can see that the auth token is no where to be found.

```
  2) test deliver/2 sends from, html and text body, subject, and headers (Bamboo.MailgunAdapterTest)
     test/lib/bamboo/adapters/mailgun_adapter_test.exs:83
     ** (Bamboo.MailgunAdapter.ApiError) There was a problem sending the email through the Mailgun API.

     Here is the response:

     %HTTPoison.Response{body: "Error!!", headers: [{"server", "Cowboy"}, {"date", "Tue, 19 Apr 2016 00:20:00 GMT"}, {"content-length", "7"}, {"cache-control", "max-age=0, private, must-revalidate"}], status_code: 500}


     Here are the params we sent:

     "from=from%40foo.com&html=HTML+BODY&subject=My+Subject&text=TEXT+BODY"

     stacktrace:
       (bamboo) lib/bamboo/adapters/mailgun_adapter.ex:52: Bamboo.MailgunAdapter.deliver/2
       test/lib/bamboo/adapters/mailgun_adapter_test.exs:91
```

princemaple:
@paulcsmith addressed the comments and rebased again. re: auth token see my inline reply :)

lib/bamboo/adapters/mailgun_adapter.ex
paulcsmith: Oh that makes perfect sense! Thanks for the example output

paulcsmith:
Perfect! I'll take a look again tomorrow and merge it in!

princemaple:
Thanks!

princemaple:
sorry, closed by accident. rebased to current master.

princemaple:
@paulcsmith let me know if there is any more work to be done.

paulcsmith:
Looks really good! Thanks again :)

PR #127, opened 2016-04-19
Show the email in the list instead of the selected

drapergeek:
This fixes a bug where the sidebar only showed the `from` address
of the currently selected email.

This also changes the factory to use a sequence so that the tests that require
multiple emails will create unique data.


paulcsmith:
Good catch! Thanks @drapergeek. I'll merge this in once CI passes

PR #133, opened 2016-04-26
Start TaskSupervisor by default

### LINKED ISSUE #131: Start Bamboo.TaskSupervisorStrategy automatically
This makes it easier to get started and this way it won't conflict if two applications both try to use TaskSupervisorDeliveryStrategy

Also raise if `TaskSupervisorDeliveryStrategy.child_spec` is called.

It might also be cool to make it so you can configure the supervisor options

```elixir
config :bamboo, :task_supervisor_delivery_strategy, max_restarts: 10, restart_strategy: :transient
```

paulcsmith:
The reason for this is that if two applications start bamboo and both of
them added Bamboo.TaskSupervisorStrategy.child_spec to their children,
then there would be an error because it was started twice.

This circumvents that by starting it just once when Bamboo starts. This
also simplifies the installation instructions.

Closes #131

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,16 +65,6 @@ config :my_app, MyApp.Mailer,
   end
   ```
 
-3. Add the the `Bamboo.TaskSupervisor` as a child to your supervisor. This is necessary for `deliver_later` to work.
-
-  ```elixir
-  # Usually in lib/my_app_name.ex
-  children = [
-    # This is where you add the supervisor that handles deliver_later calls
-    Bamboo.TaskSupervisorStrategy.child_spec
-  ]
-  ```
-
 ## Getting Started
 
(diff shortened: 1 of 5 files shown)


princemaple:
an unrelated note: should Bamboo.SentEmail always be there?
You could leave this to the user to start in test cases..?

paulcsmith:
That is a good question. I think that it's ok to start it here though since the sent email adapter could be used on staging for example. It's also such little overhead to start a supervisor that it's probably the best option for making bamboo simple to get started with. 

However if this ever becomes a problem then I would happily reconsider!

drapergeek:
I like it! LGTM

paulcsmith:
Thanks for the reviews and comments @drapergeek and @princemaple

Issue #60, opened 2016-02-10
Add SMTPAdapter


paulcsmith:
Possible libraries to use

* https://github.com/nurugger07/hermes
* https://github.com/antp/mailer

aselder:
I'm really looking forward to this adapter, since we use AWS's SMTP servers. I may try and start hacking on this. Anybody want to help?

paulcsmith:
@aselder That would be really great. I haven't needed SMTP so it would be awesome if you worked on it in a real world scenario. I think @BlakeWilliams started work on it, but I'm not sure how far he got

paulcsmith:
@aselder How are things going with the SMTP adapter. If you need any help feel free to ping me :)

lpil:
:+1: I would love to be able to use this lib with AWS's SMTP service.

paulcsmith:
@anellis This is a trickier issue, but it would be *super* helpful to have in Bamboo. If you have the time, creating an SMTP adapter with https://github.com/antp/mailer or https://github.com/nurugger07/hermes would be super awesome.

I'd be happy to help with anything you need. Here are some docs for getting started: https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapter.ex

keichan34:
I'd like to see this too -- I'm currently using `mailer` but this looks like it'd be easier to use in my case. My 2¢ would be to not use `mailer` as the SMTP backend but instead use `gen_smtp` (`mailer` uses `gen_smtp` behind the scenes anyways) or hermes (although I can't provide any advice on that since I haven't used it before).

kdisneur:
I started a PullRequest, using `mailer`, any help/feedbacks would be very welcome: https://github.com/thoughtbot/bamboo/pull/135 :)

@keichan34 I looked at this issue yesterday and your comment was not present yet. It's a shame because you seem right, using `gen_smtp` would be a good idea:

1. it would allow people to use `mailer` with a different configuration, if needed by another part of the application
2. It would allow to add features more easily into the `bamboo` adapter
3. `mailer` use its own templating system, so we need call more internal stuff. So, I guess it makes the current implementation more brittle

paulcsmith:
An SMTPAdapter can be found at: https://github.com/fewlinesco/bamboo_smtp Thanks @kdisneur!

Feel free to give the SMTPAdapter a shot and LMK how it goes :) @keichan34 @lpil and @aselder

PR #135, opened 2016-04-26
Add an SMTP adapter

kdisneur:
The current implementation supports:

* sending emails through SMTP (use `gen_smtp`)
* `tls`, and `ssl` protocols
* `cc` and `bcc`

I'm not quite sure if it meets `bamboo` standards but it works on my application. I just wanted to create the PR early to have feedbacks as soon as possible.

We still need to add, at least:

* [x] tests

Any feedback is welcome 😸

### DIFF
--- a/lib/bamboo/adapters/smtp_adapter.ex
+++ b/lib/bamboo/adapters/smtp_adapter.ex
@@ -82,6 +82,14 @@ defmodule Bamboo.SMTPAdapter do
     add_smtp_body_line(body, :cc, format_email(recipients, :cc))
   end
 
+  defp add_custom_header(body, {key, value}) do
+    body <> key <> ": " <> value <> "\r\n"
+  end
+
+  defp add_custom_headers(body, %Bamboo.Email{headers: headers}) do
+    Enum.reduce(headers, body, &add_custom_header(&2, &1))
+  end
+
   defp add_ending_header(body) do
     body <> "\r\n"
   end
@@ -157,6 +165,7 @@ defmodule Bamboo.SMTPAdapter do
     |> add_bcc(email)
     |> add_cc(email)
     |> add_to(email)
+    |> add_custom_headers(email)
     |> add_mime_header
     |> add_multipart_header(multi_part_delimiter)
     |> add_ending_header
--- a/lib/bamboo/adapters/smtp_adapter.ex
+++ b/lib/bamboo/adapters/smtp_adapter.ex
@@ -75,15 +75,15 @@ defmodule Bamboo.SMTPAdapter do
   end
 
   defp add_bcc(body, %Bamboo.Email{bcc: recipients}) do
-    add_smtp_body_line(body, :bcc, format_email(recipients, :bcc))
+    add_smtp_header_line(body, :bcc, format_email(recipients, :bcc))
   end
 
   defp add_cc(body, %Bamboo.Email{cc: recipients}) do
-    add_smtp_body_line(body, :cc, format_email(recipients, :cc))
+    add_smtp_header_line(body, :cc, format_email(recipients, :cc))
   end
 
   defp add_custom_header(body, {key, value}) do
-    body <> key <> ": " <> value <> "\r\n"
+    add_smtp_header_line(body, key, value)
   end
 
   defp add_custom_headers(body, %Bamboo.Email{headers: headers}) do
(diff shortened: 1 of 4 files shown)


princemaple:
Since Bamboo requires [Elixir ~> 1.2](https://github.com/thoughtbot/bamboo/blob/master/mix.exs#L9), I'd say swap `Dict`s with `Map`s?

@paulcsmith Also, this feature adds so much dependency, would you / thoughtbot be interested in keeping it as a **separate repo / hex package** that you guys officially supports / maintains.

kdisneur:
As suggested by @keichan34 in this issue https://github.com/thoughtbot/bamboo/issues/60, I'm reworking it to use `gen_smtp` so it would reduce the number of dependencies.

I used `Dict`s instead of `Map`s to make it easier to talk with `mailer` so moving to `gen_smtp` will also get rid of this. 👍 

I created as a pull request as #60 seemed to suggest they wanted it here. Instead, creating it as a separate hex repo / package makes total sense for me as Bamboo has been created to be extensible. :)

princemaple:
All good points. Thanks Kevin!

kdisneur:
I just pushed a new commit replacing `mailer` by `gen_smtp`.

It's currently working as plain text.

I still need to implement the `HTML` emails 😁 

I have to stop working on this today so will try to finish the HTML part tonight.

kdisneur:
Just before starting my working day \o. 
I just added the HTML support. So if people want to try it, hopefully it should work.

I still need to find the best way to generate the multipart delimiter. If you have any suggestions :)

princemaple:
I should be able to help test your implementation tomorrow morning (Australia time)
I'll update you if I do get to it.

lib/bamboo/adapters/smtp_adapter.ex
```diff
@@ -0,0 +1,265 @@
+defmodule Bamboo.SMTPAdapter do
+  @moduledoc """
+  Sends email using SMTP protocol.
+
+  Use this adapter to send emails through SMTP protocol. Requires that some
```
drapergeek: The word "protocol" can be removed here, it's redundant.

drapergeek: How about "This adapter requires that"

lib/bamboo/adapters/smtp_adapter.ex
```diff
@@ -0,0 +1,265 @@
...
+        adapter: Bamboo.SMTPAdapter,
+        server: "smtp.domain",
+        port: 1025,
+        username: "your.name@your.domain"
+        password: "pa55word",
+        tls: :if_available, # can be `:always` or `:never`
+        ssl: :false, # can be `:true`
+        retries: 1
+
+      # Define a Mailer. Maybe in lib/my_app/mailer.ex
```
drapergeek: I think this is defined elsewhere, does it need to be repeated here?

lib/bamboo/adapters/smtp_adapter.ex
```diff
@@ -0,0 +1,265 @@
...
+  end
+
+  @doc false
+  def handle_config(config) do
+    config
+    |> check_required_configuration
+    |> put_default_configuration
+  end
+
+  defp handle_response({:error, reason, detail}) do
```
drapergeek: What do you think about grouping this function with the `deliver` where it is used?

anellis:
@kdisneur I started on this as well a few days ago and got a pretty good test situation set up, if you'd like use my initial tests I've pushed it up for you [here](https://github.com/thoughtbot/bamboo/compare/smtp_adapter?expand=1). I tried to follow the same pattern as the others in doing a fake server type of situation

paulcsmith:
@kdisneur First off, thank you so much! An SMTP adapter will be very helpful for a **lot** of people.

I think it's fine to have the PR here so we can help with the review, but I think it would be best to extract this to it's own package and repo once review is done. The reason for that is that I don't think it makes sense to add `gen_smtp` as a dependency to Bamboo for those not using SMTP, and because the adapters that ship with Bamboo are ones we have tried or are currently using. Because we're using them we know they work. The last reason is that once Bamboo hits 1.0 I will move all adapters (except for LocalAdapter and TestAdapter) to their own repos and Bamboo will just have the core functionality. Users can then include whatever adapter they would like.

So I think we can review things here and help you out with tests if you'd like and then once it looks good, could you make a new repo and hex package for it? Then you can ping me and I'll add a link to it in the repo just like I did for the SparkpostAdapter https://github.com/thoughtbot/bamboo/commit/04df26bca295949a11047fcc8efe1c29c72321f7

kdisneur:
Thanks @princemaple, it would be awesome if you have some ideas of how to test them. I will not have time to look at this before tomorrow morning (French time). @anellis suggested to look at [his tests](https://github.com/thoughtbot/bamboo/compare/smtp_adapter?expand=1). It can maybe help.

Thanks @drapergeek for the review. I will update that tomorrow morning.

Thanks @anellis, I will have a look to your tests tomorrow morning. That could be really helpful. 🙇 

Thanks @paulcsmith, I will create a new repository / hex package as soon as the PR is done. It makes total sense to extract the adapters to other repositories.

princemaple:
tested with a newly _mixed_ app, with mailgun smtp

```elixir
config :test_bamboo_smtp, TestBambooSMTP.Mailer,
  adapter: Bamboo.SMTPAdapter,
  server: "smtp.mailgun.org",
  port: 2525,
  username: "postmaster@MY_SECRET_SANDBOX.mailgun.org",
  password: "PASSWORD",
  tls: :if_available, # can be `:always` or `:never`
  ssl: :false, # can be `:true`
  retries: 1

defmodule TestBambooSMTP.Mailer do
  use Bamboo.Mailer, otp_app: :test_bamboo_smtp
end

defmodule TestBambooSMTP do
  use Application

  def start(_type, _arg) do
    Supervisor.start_link([], [strategy: :one_for_one])
  end
end
```

```elixir
iex(3)> TestBambooSMTP.Mailer.deliver_now %Bamboo.Email{from: "pochen@somewhere.com", to: "chenpaul914@gmail.com", subject: "test smtp", html_body: "hi", text_body: "hi"}

13:53:52.412 [debug] Sending email with Bamboo.SMTPAdapter:

%Bamboo.Email{assigns: %{}, bcc: [], cc: [], from: {nil, "pochen@somewhere.com"}, headers: %{}, html_body: "hi", private: %{}, subject: "test smtp", text_body: "hi", to: [nil: "chenpaul914@gmail.com"]}

%Bamboo.Email{assigns: %{}, bcc: [], cc: [],
 from: {nil, "pochen@somewhere.com"}, headers: %{}, html_body: "hi",
 private: %{}, subject: "test smtp", text_body: "hi",
 to: [nil: "chenpaul914@gmail.com"]}
```

Mailgun logs screenshot
screenshot 2016-04-27

Inbox screenshot
screenshot 2016-04-27 1

The datetimes in the screenshots differ, but it's just mailgun web interface not respecting my local time :)

@kdisneur I've read through the test written by @anellis , I think it is a good way of testing the SMTP adapter, i.e. __not__ to test the smtp client. After all we just need to know whether our adapters correctly transformed the data into the right formats; we don't want to test whether mailgun sends the email, whether httpoison makes the http call, or whether gen_smtp does the right thing.

lib/bamboo/adapters/smtp_adapter.ex
kdisneur: I added this Mailer example because it's repeated in all adapters so I just wanted to stay consistent with other adapters documentation.

lib/bamboo/adapters/smtp_adapter.ex
kdisneur: If we group `handle_response` with `deliver`, we would have: 

```elixir
defmodule SMTPAdapter do
  def deliver(...) do
    # ...
  end

  defp handle_response(...) do
    # ...
  end

  def handle_config(...) do
    # ...
  end
end
```

I believe it's harder for someone to detect what's the module interface if `public` and `private` functions are mixed together. It's not a problem with `Erlang` because you define your exported functions at the top of the file but with `Elixir` the only clue is `def` vs `defp`, so mixing them would make other developers life harder isn't it?

I'm fairly new in `Elixir` (and functional programming) so I'm probably totally wrong. Maybe developers are used to mix public/private together. Anyway, I would love to hear your thinking about this.

lib/bamboo/adapters/smtp_adapter.ex
kdisneur: 👍

kdisneur:
Thank you very much for your test and feedback @princemaple 

I agree, it makes no sense to test `gen_smtp` behaviour. So, I will have a look to @anellis work and start to write my tests with a fake `gen_smtp` adapter, to be able to test only the bamboo SMTPAdapter.

I don't think I will have time to finish this part before my day work. So, I will more likely give you some updates tomorrow morning.

lib/bamboo/adapters/smtp_adapter.ex
paulcsmith: I don't think there's a hard and fast rule, but I don't mind mixing them under certain circumstances. 

If the private function is called by just one public function, I'll put the private function right below it. If the private function is called by multiple public functions I usually put it at the bottom. I think either way you go here is ok

paulcsmith:
Sounds good! No rush in anything. I really appreciate your work in this! And thanks @princemaple for testing out the adapter :)

paulcsmith:
@princemaple did the adapter send the email twice or did you send it twice? I'm on my phone so maybe I missed something. Code snippets are a bit tricky to read on a small screen :P

lib/bamboo/adapters/smtp_adapter.ex
drapergeek: Sounds good, just a thought!

princemaple:
@paulcsmith it worked as expected, i.e. sent once. The code snippet was a little messed up, copying from iex to here mixes up the logging and the returned value.

lib/bamboo/adapters/smtp_adapter.ex
paulcsmith: Oh and that's a good point about figuring out what is exported and what's not, but I think most people will usually check the docs and not the source code itself. Also, docs tags are usually added to public functions in libraries which makes it easier to find them. I hope that helps :)

paulcsmith:
Oh that makes sense! Thanks for the clarification.

kdisneur:
I added some tests, I guess they could look better but at leat they test the behaviour. If you have any suggestion to make them look better. :)

@paulcsmith : any conventions about how we should name bamboo adapters? I can see the only one listed in the readme is named `bamboo_sparkpost` so, I would suggest `bamboo_smtp`.

kdisneur:
After looking at hex.pm it seems like the convention is to use `bamboo_<adapter-name>` so I created the package: `bamboo_smtp`: https://hex.pm/packages/bamboo_smtp

Thanks everyone for the help with this PullRequest. If you have any more feedback, please create an issue on: https://github.com/fewlinesco/bamboo_smtp

PR #136, opened 2016-04-26
Remove process_name. Add shared mode

### LINKED ISSUE #130: Change Bamboo.Test, process_name: :foo, to Bamboo.Test, shared: true
This makes it so that people don't have to come up with a process name and it is a bit more clear about when it is used and why it is used.

- [x] Change config from `process_name: :whatever` to `shared: true`
- [x] Store `test_process_name` as `self()`, and remove the line for registering the process: https://github.com/thoughtbot/bamboo/blob/c46485e67884565338788e010fd8577c1e755aa7/lib/bamboo/test.ex#L125

This should simplify the docs, simplify the code, and make things more clear. Thanks to @josevalim for pointing out that this could be improved

paulcsmith:
Closes #130. This makes usage simpler and adds some clarity to what is
going on.

### DIFF
--- a/lib/bamboo/test.ex
+++ b/lib/bamboo/test.ex
@@ -13,10 +13,10 @@ defmodule Bamboo.Test do
   ## Note on sending from other processes
 
   If you are sending emails from another process (for example, from inside a
-  Task or GenServer) you may need to use the `process_name` option when using
+  Task or GenServer) you may need to use shared mode when using
   `Bamboo.Test`. See the docs `__using__/1` for an example.
 
-  For most scenarios you will not need the `process_name` option.
+  For most scenarios you will not need shared mode.
 
   ## In your config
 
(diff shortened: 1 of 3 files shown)


lib/bamboo/test.ex
```diff
@@ -82,31 +82,28 @@ defmodule Bamboo.Test do
...
   headless browser like phantomjs.
   """
-  defmacro __using__(process_name: process_name) do
+  defmacro __using__(shared: true) do
     quote do
       setup tags do
         if tags[:async] do
           raise """
-          passing process_name to Bamboo.Test cannot be done for async tests.
+          using shared mode with Bamboo.Test cannot be done for async tests.
```
jsteiner: Should `using` be capitalized?

BlakeWilliams: What do you think of recommending removing `shared: true` from the `use`?

eg:

```
Replace `use Bamboo.test, shared: true` with `use Bamboo.test`
```

jsteiner: Also, thoughts on changing this from the [passive voice](https://en.wikipedia.org/wiki/English_passive_voice) to make it a little more clear? Maybe something like `You cannot use Bamboo.Test shared mode with async tests`?

BlakeWilliams: I'm a fan, I agree that it makes it more clear.

drapergeek:
LGTM

I appreciate the explanations in the erorrs.

lib/bamboo/test.ex
```diff
@@ -115,14 +112,13 @@ defmodule Bamboo.Test do
                delivering from within Task.async or Process.spawn), try using
                Mailer.deliver_later. If you use Mailer.deliver_later without
                spawning another process you can use Bamboo.Test with [async:
-               true] and without the process_name option.
-            3) If you are doing an acceptance test that requires the process_name
-               option, try using a controller test instead. Then see if the test
-               works without the process_name option.
+               true] and without the shared mode.
+            3) If you are doing an acceptance test that requires shared mode,
```
BlakeWilliams: Replacing `doing` with "writing", "running", or "writing/running" maybe?

lib/bamboo/test.ex
paulcsmith: Yes it should be capitalized. Thank you! I think that recommending once when the email fails to deliver and then again when you use shared mode with async true would probably be more confusing/annoying than helpful.

lib/bamboo/test.ex
BlakeWilliams: Re-reading, I think writing makes sense.

BlakeWilliams:
👍 

I'm a huge fan of how descriptive the errors are!

jsteiner:
👍

lib/bamboo/test.ex
paulcsmith: Oh never mind that should not be capitalized. The beginning of errors are supposed to be lowercased because the sentence starts with `raised RuntimeError` or something like that. Here's an example: https://github.com/phoenixframework/phoenix/blob/e3c23a364af4f2283e2498fdb47c60161f262a1f/lib/phoenix/router.ex#L432

lib/bamboo/test.ex
paulcsmith: I like writing. Great suggestion!

paulcsmith:
Thank you for the reviews :)

lib/bamboo/test.ex
paulcsmith: I like this, but missed it during your review! I’ll add it in later

PR #137, opened 2016-04-26
Update docs in Bamboo.Phoenix

paulcsmith:
After looking through this I think "template" might be the more correct word here

### DIFF
--- a/lib/bamboo/phoenix.ex
+++ b/lib/bamboo/phoenix.ex
@@ -15,10 +15,10 @@ defmodule Bamboo.Phoenix do
         def text_and_html_email_with_layout do
           new_email()
           # You could set just a text layout or just an html layout
-          |> put_text_layout
-          |> put_html_layout
+          |> put_text_layout({MyApp.LayoutView, "email.text"})
+          |> put_html_layout({MyApp.LayoutView, "email.html"})
           # Or you can set a layout for both html and text at the same time
-          |> put_layout
+          |> put_layout({MyApp.LayoutView, :email})
           # Pass an atom to render html AND plain text templates
           |> render(:text_and_html_email)
         end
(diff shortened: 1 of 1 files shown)


lib/bamboo/phoenix.ex
```diff
@@ -15,10 +15,10 @@ defmodule Bamboo.Phoenix do
         def text_and_html_email_with_layout do
           new_email()
           # You could set just a text layout or just an html layout
-          |> put_text_layout
-          |> put_html_layout
+          |> put_text_layout({MyApp.LayoutView, "email.text"})
+          |> put_html_layout({MyApp.LayoutView, "email.html"})
```
paulcsmith: I think adding the arguments might make things clearer

lib/bamboo/phoenix.ex
```diff
@@ -122,7 +122,7 @@ defmodule Bamboo.Phoenix do
 
       def text_email_layout do
         new_email
-        # Will use MyApp.LayoutView with email.text layout when rendering text emails
+        # Will use MyApp.LayoutView with email.text template when rendering text emails
```
paulcsmith: I'm not sure about this, but I think template is probably more correct. Thoughts?

anellis: Hmm I keep going back and forth on this one. In the method definitions though we have
`put_text_layout(email, layout)`, so right now in a variety of places we're referring to that `email.text` or `email.html` as a layout. So, idk. I'd be fine with referring to it as a template but if so I think we need to change it everywhere.

anellis: Actually, sorry, just saw that its actually broken down to `{layout, template}`, so yes, 👍  on the change

anellis:
LGTM

paulcsmith:
Thanks for the review!

PR #150, opened 2016-05-02
Add types and specs

cloud8421:
Hello, I've added some initial types and specs (excluding the Plug/Phoenix side of things).

Incidentally, I've been using the library for a few weeks now (just a to generate a daily digest to myself) and it works very nicely, I'm really happy about it! Nice job.

### DIFF
--- a/lib/bamboo/email.ex
+++ b/lib/bamboo/email.ex
@@ -150,7 +150,7 @@ defmodule Bamboo.Email do
 
   ## Examples
 
-    Bamboo.Email.get_address({"Paul", "paul@thoughtbot.com"}) # "paul@thoughtbot.com"
+      Bamboo.Email.get_address({"Paul", "paul@thoughtbot.com"}) # "paul@thoughtbot.com"
   """
   @spec get_address(address) :: String.t | no_return
   def get_address({_name, address}), do: address
--- a/lib/bamboo/email.ex
+++ b/lib/bamboo/email.ex
@@ -70,9 +70,9 @@ defmodule Bamboo.Email do
   @type address :: String.t | {String.t, String.t}
 
   @type t :: %__MODULE__{
-    to: nil | address | [address],
-    cc: nil | address | [address],
-    bcc: nil | address | [address],
+    to: nil | any,
+    cc: nil | any,
+    bcc: nil | any,
     subject: nil | String.t,
     html_body: nil | String.t,
     text_body: nil | String.t,
@@ -183,7 +183,7 @@ defmodule Bamboo.Email do
 
       put_private(email, :tags, "welcome-email")
   """
-  @spec put_private(__MODULE__.t, atom, String.t) :: __MODULE__.t
+  @spec put_private(__MODULE__.t, atom, any) :: __MODULE__.t
   def put_private(%Email{private: private} = email, key, value) do
     %{email | private: Map.put(private, key, value)}
   end
--- a/lib/bamboo/email.ex
+++ b/lib/bamboo/email.ex
@@ -117,7 +117,6 @@ defmodule Bamboo.Email do
     @doc """
     Sets the #{function_name} on the email
     """
-    @spec unquote(function_name)(__MODULE__.t, String.t) :: __MODULE__.t
     def unquote(function_name)(email, attr) do
       Map.put(email, unquote(function_name), attr)
     end
--- a/lib/bamboo/email.ex
+++ b/lib/bamboo/email.ex
@@ -67,6 +67,20 @@ defmodule Bamboo.Email do
   ```
   """
 
+  @type address :: String.t | {String.t, String.t}
+
+  @type t :: %__MODULE__{
+    to: nil | address | [address],
+    cc: nil | address | [address],
+    bcc: nil | address | [address],
+    subject: nil | String.t,
+    html_body: nil | String.t,
+    text_body: nil | String.t,
+    headers: %{String.t => String.t},
+    assigns: %{atom => any},
+    private: %{atom => any}
+  }
+
   defstruct from: nil,
       to: nil,
       cc: nil,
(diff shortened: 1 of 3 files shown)


lib/bamboo/formatter.ex
```diff
@@ -58,6 +58,8 @@ defprotocol Bamboo.Formatter do
   `:from`, `:to`, `:cc` or `:bcc`. You can pattern match on this to customize
   the address.
   """
+  @type opts :: %{type: :from | :to | :cc | :bcc}
```
princemaple: a newline after this, maybe, for readability?

princemaple:
@paulcsmith have you missed this goodie?

princemaple:
@cloud8421 out of curiosity, how did you generate your daily digest? did you use any task queue / cron?

paulcsmith:
Did not miss it, just been busy in client work :D

Thanks so much for this @cloud8421. I'm glad you're liking it so far! I'll have more time to review this later on in the week :)

@princemaple FWIW, we do a daily digest and use Heroku Scheduler with a mix task and it works wonderfully, but you need to be on Heroku. Not sure if that helps, but hopefully it does!

princemaple:
@paulcsmith thanks for the information :)
I've been searching for a elixir/erlang solution. Reason being it would be much easier to deploy as docker images, as I wouldn't have to touch any lower-level system tools, e.g. cron, or break basic principles like one process/command per container. I also have this requirement that the solution needs to be able to run as the unique process across an Elixir / Phoenix cluster, so the jobs created by it wouldn't repeat. Since I haven't found one, I've started looking into `:global` and writing one of my own...

paulcsmith:
With those requirements you Kay be interested in using a package called Quantum

It's basically cron, but pure elixir. We have used it on a client project and it is working well so far. You can even set schedules with human readable language which is nice, e.g. "@hourly"

On phone so don't have a link but it's on hex :)

princemaple:
Thanks, that one is interesting.

cloud8421:
@paulcsmith: sounds great!

@princemaple:

The project I’m working on is a system that gathers information in the background and reacts to certain events.

- a clock process ticks every second
- a scheduler consumer/producer processes every tick, generating minute, hour and day events
- a notification consumer listens to all “hour” events, discarding all except 7am, where it fires the email (this is done via simply pattern matching on the events coming in)

Both consumer and producers are `GenServer` processes that bind themselves to a `Phoenix.PubSub` instance.

Everything runs in the same application. Hope it's clear!

princemaple:
@cloud8421 yep, that's clear enough :) thank you!

lib/bamboo/email.ex
```diff
@@ -67,6 +67,20 @@ defmodule Bamboo.Email do
   ```
   """
 
+  @type address :: String.t | {String.t, String.t}
+
+  @type t :: %__MODULE__{
+    to: nil | address | [address],
+    cc: nil | address | [address],
+    bcc: nil | address | [address],
```
paulcsmith: I think this should be `nil | any` because the email address can technically be whatever you want as long as it defines the `Bamboo.Formatter` protocol

lib/bamboo/email.ex
```diff
@@ -94,6 +108,7 @@ defmodule Bamboo.Email do
       # Same as %Bamboo.Email{from: "support@myapp.com"}
       new_email(from: "support@myapp.com")
   """
+  @spec new_email(Enum.t) :: __MODULE__.t
```
paulcsmith: I believe it can accept `map, Enum.t` since those are what `struct!` accepts

lib/bamboo/email.ex
```diff
@@ -102,6 +117,7 @@ defmodule Bamboo.Email do
     @doc """
     Sets the #{function_name} on the email
     """
+    @spec unquote(function_name)(__MODULE__.t, String.t) :: __MODULE__.t
```
paulcsmith: This one is tricky so we may want to remove it for now. For example, from, to, cc, bcc accept `any` instead of `String.t`.

So maybe we should have individual functions from, to, subject, etc. with their own type specs. That may be outside the scope of this PR so maybe just removing this spec for now is ok

lib/bamboo/email.ex
```diff
@@ -164,6 +183,7 @@ defmodule Bamboo.Email do
 
       put_private(email, :tags, "welcome-email")
   """
+  @spec put_private(__MODULE__.t, atom, String.t) :: __MODULE__.t
```
paulcsmith: Rather than `String.t` it should be `any` because you can put whatever you like in the private key

lib/bamboo/formatter.ex
paulcsmith: Yes I think this would make it read a bit better

paulcsmith:
A few comments and then this will be ready to merge. Thanks @cloud8421

lib/bamboo/email.ex
cloud8421: Of course, that's true.

cloud8421:
@paulcsmith will make the changes and push. Is there a policy around squashing/rebasing?

paulcsmith:
@cloud8421 I typically squash the commits before merging, but I can't easily rebase it from the web. So if you can rebase I can manage squashing/merging. Does that work for you?

paulcsmith:
@cloud8421 Just checking in to see if you need any help with this :)

cloud8421:
Yes I think I'll be able to make those changes later today

Claudio

---

lib/bamboo/email.ex
cloud8421: Isn't that the second argument? (pasting here the original spec)

```
struct!(module | map, Enum.t) :: map | no_return
```

cloud8421:
@paulcsmith rebased, made needed changes and left a question :)

lib/bamboo/email.ex
paulcsmith: You are correct! Good catch

paulcsmith:
This looks great. Thanks @cloud8421!

Issue #53, opened 2016-02-08
Add ability to add attachments


paulcsmith:
This is probably going to come a bit later since I have never needed to send an attachment in an email, it's always a link back to the app. I know this is useful to people though so it should be added.

hdcos:
yes please, my business needs it, and currently mailgun works but doesn't support utf8 so cannot put it in production

michalvalasek:
Any news on this issue guys?

davekapp:
We would definitely like to take advantage of this as well, so another +1 from me.

pradhyu:
This would be a great feature to add

paulcsmith:
Closed by #156

paulcsmith:
Coming in the 1.0rc1 that I will try to push to hex today

davekapp:
That's awesome! Thank you so much!

ijunaid8989:
@paulcsmith is there any inline attachment options available?

PR #156, opened 2016-05-11
Add attachment support

andrewtimberlake:
This adds support for attachments, see #53 
I have added support in an adapter at https://github.com/andrewtimberlake/bamboo_sparkpost/tree/feature/attachments


paulcsmith:
@andrewtimberlake I'll try to take a look at this tomorrow. Thanks for giving this a shot! If I don't respond in a few days please ping me! I've been a bit busy and it might slip my mind

andrewtimberlake:
@paulcsmith ping, because you asked for it :-)

paulcsmith:
Hey @andrewtimberlake! Thanks for the ping. I think this should also support sending in the Base64 encoded file. Sometimes that's the easiest thing to do if the attachment is stored elsewhere. What do you think about that? 

I also think that in the Mailer the attachment should be normalized before it's sent to the adapter. So instead of Base64 encoding it in the adapter, it's one when you deliver the email, right before going to the adapter. That way the adapter has just one type of attachment it needs to handle. Does that make sense?

andrewtimberlake:
I had thought about adding an attachment with raw content would be useful.
One of the reasons I didn’t do any of the base64 encoding in Bamboo itself was to prevent copying large amounts of data when switching processes, like `deliver_later`
Is there any concern about different adapters needing to deal with the data differently? SparkPost needs the data base64 encoded without line endings. I could strip any line endings but I’m not sure if any other issues could crop up.

hisea:
Any follow up on this PR? I just opened PR #179  to add support attachment for Mandrill adapter. Please ping me if this PR gets changed and adjustment is needed in PR #179.

paulcsmith:
@andrewtimberlake and @hisea So I've been thinking about this for awhile (sorry for the late response) and I think this looks good, but I would like to hold off on it until after 1.0 is out. The reason for that is that when we add attachment support I'd also like to make sure all (or at least *most*) adapters support it, and that we've had a chance to test them out. 

For now I'd like to focus on getting the core of Bamboo ready to go and hit 1.0, then we can add on the attachment stuff once 1.0 is shipped. So I'll leave this open and address PR #179 once 1.0 is shipped! This might be a number of months, but I appreciate having this open so that other people can give it a try or leave comments!

Thanks for your work guys!

hisea:
👍

lib/bamboo/attachment.ex
```diff
@@ -0,0 +1,34 @@
...
+  Examples:
+    Attachment.new("/path/to/attachment.png")
+    Attachment.new("/path/to/attachment.png", filename: "image.png")
+    Attachment.new("/path/to/attachment.png", filename: "image.png", content_type: "image/png")
+    Attachment.new(params["file"]) # Where params["file"] is a %Plug.Upload
+  """
+  def new(path, opts \\ [])
+  if Code.ensure_loaded?(Plug) do
+    def new(%Plug.Upload{filename: filename, content_type: content_type, path: path}, opts), do:
+      new(path, Dict.merge([filename: filename, content_type: content_type], opts))
```
princemaple: `Dict` -> `Keyword`

suhrawardi:
Look fwd to have this merged in... 🙂

benjamin79:
Why no merge?

paulcsmith:
@benjamin79 Good question. I'm preparing for Bamboo 1.0 right now and hope to get this in as I get closer. The feature itself and the code is great, but I also need to make sure that I add documentation to the other adapters that don't yet support attachments. I don't want people to see this and then be confused about why attachments aren't working for the adapter they are using.

I hope to tackle this in the next few weeks

samrispaud:
What's the status of this? Excited for this to get merged

paulcsmith:
I want to release this even though not all adapters have implemented attachment support, but I also don't want people to get weird and unexpected errors when trying to use attachments with adapters that don't use it. I think I've found something that will work. I'll try it out, then merge this in to get ready for 1.0. We're doing some special projects at thoughtbot right now so I likely won't be able to get to this until 1-2 weeks. Maybe sooner if I can find some time after work!

paulcsmith:
Thanks @andrewtimberlake! A lot of people are really excited about this and greatly appreciate your work

Issue #154, opened 2016-05-11
Strip assigns from the email when testing


paulcsmith:
We should strip `assigns` (or set it to something like `:assigns_removed_for_testing`). The reason for this is that the assigns are not important to the email, they only matter for rendering the html/text body. So as long as the bodies are right it's fine.

The problem is that sometimes the assigns don't exactly match if you are preloading in the test setup, or *not* preloading in the test setup. It makes it so that the assigns may differ in the sent email, even though the sent email is sending what you expect. So stripping the assigns is a simple way to get this working

1) Set email.assigns to `:assigns_removed_for_testing` when sending from Bamboo.TestAdapter (https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/test_adapter.ex#L24)
2) When asserting, the email should be normalized (as it already is doing), and then should set assigns to `:assigns_removed_for_testing` so it matches what the TestAdapter sends

There should probably be just one function that handles this. Like `Bamboo.TestAdapter.strip_assigns` or `Bamboo.TestAdapter.clean_assigns`. I think I like `clean_assigns` best. There should not be any documentation for that function since it's not meant to be used outside of Bamboo.Test and Bamboo.TestAdapter

PR #158, opened 2016-05-14
Strip assigns from the email when testing

andrewtimberlake:
I needed this because I’m passing `conn` into some templates
Implements #154

### DIFF
--- a/test/lib/bamboo/adapters/test_adapter_test.exs
+++ b/test/lib/bamboo/adapters/test_adapter_test.exs
@@ -160,15 +160,11 @@ defmodule Bamboo.TestAdapterTest do
     assert_delivered_email sent_email
   end
 
-  # The actual sent email and the email tested may have differences in their assigns
-  # Even if the resulting body is the same
-  test "assertion helpers remove assigns" do
-    test_email = new_email(from: "foo@bar.com", to: "bar@baz.com")
-    sent_email = %{test_email | assigns: %{foo: :bar}}
-    expected_email = %{test_email | assigns: %{foo: :baz}}
+  test "delivered emails have normalized assigns" do
+    email = new_email(from: "foo@bar.com", to: "bar@baz.com", assigns: :anything)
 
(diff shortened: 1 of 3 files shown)


test/lib/bamboo/adapters/test_adapter_test.exs
```diff
@@ -157,4 +159,16 @@ defmodule Bamboo.TestAdapterTest do
 
     assert_delivered_email sent_email
   end
+
+  # The actual sent email and the email tested may have differences in their assigns
+  # Even if the resulting body is the same
+  test "assertion helpers remove assigns" do
```
paulcsmith: I think the comment can be removed and maybe the test can say something like `"delivered emails have normalized assigns"` or something like that. What do you think?

test/lib/bamboo/adapters/test_adapter_test.exs
```diff
@@ -157,4 +159,16 @@ defmodule Bamboo.TestAdapterTest do
...
+    test_email = new_email(from: "foo@bar.com", to: "bar@baz.com")
+    sent_email = %{test_email | assigns: %{foo: :bar}}
+    expected_email = %{test_email | assigns: %{foo: :baz}}
+
+    sent_email |> TestMailer.deliver_now
+
+    assert_delivered_email expected_email
```
paulcsmith: I think this test could be simplified a bit by doing something like

```elixir
email = new_email(from: "foo@bar.com", to: "bar@baz.com", assigns: :anything)

email |> TestMailer.deliver_now

assert_delivered_email %{email | assigns: :assigns_removed_for_testing}
```

That way the assertion more clearly shows the expected result. What do you think?

paulcsmith:
Two small comments in the test. The rest looks perfect. Thank you very much!

andrewtimberlake:
Good comments, I’ve implemented both changes.

paulcsmith:
Thank you so much :D

PR #163, opened 2016-05-19
Add SendGrid template support.

rafroehlich2:
The current implementation only allows using substitution tags on a specified template. This can be used as a jumping off point for future use of the `x-smtpapi` field.

### DIFF
--- a/lib/bamboo/adapters/sendgrid_adapter.ex
+++ b/lib/bamboo/adapters/sendgrid_adapter.ex
@@ -162,7 +162,9 @@ defmodule Bamboo.SendgridAdapter do
     # to neglect doing so themselves.
     case {email.text_body, email.html_body} do
       {nil, nil} -> put_text_body(body, %Email{email | text_body: " "})
-      _ -> body
+      _ -> raise """
+      cannot set the body and use a SendGrid template at the same time. Please remove the body or SendGrid template
+      """
     end |> Map.put("x-smtpapi", Poison.encode!(fields))
   end
   defp maybe_put_x_smtp_api(body, _), do: body
--- a/lib/bamboo/adapters/sendgrid_adapter.ex
+++ b/lib/bamboo/adapters/sendgrid_adapter.ex
@@ -115,6 +115,7 @@ defmodule Bamboo.SendgridAdapter do
     |> put_subject(email)
     |> put_html_body(email)
     |> put_text_body(email)
+    |> put_x_smtp_api(email)
   end
 
   defp put_from(body, %Email{from: {"", address}}), do: Map.put(body, :from, address)
@@ -154,6 +155,17 @@ defmodule Bamboo.SendgridAdapter do
 
   defp put_text_body(body, %Email{text_body: nil}), do: body
   defp put_text_body(body, %Email{text_body: text_body}), do: Map.put(body, :text, text_body)
+
+  defp put_x_smtp_api(body, %Email{private: %{"x-smtpapi" => fields}} = email) do
+    # SendGrid will error with empty bodies, even while using templates.
+    # Sets a default `text_body` if one is not specified, allowing the consumer
+    # to negelect doing so themselves.
+    case {email.text_body, email.html_body} do
+      {nil, nil} -> put_text_body(body, %Email{email | text_body: " "})
+      _ -> body
+    end |> Map.put("x-smtpapi", Poison.encode!(fields))
+  end
+  defp put_x_smtp_api(body, _), do: body
 
   defp put_addresses(body, field, addresses), do: Map.put(body, field, addresses)
   defp put_names(body, field, names) do
--- a/lib/bamboo/adapters/sendgrid_adapter.ex
+++ b/lib/bamboo/adapters/sendgrid_adapter.ex
@@ -158,14 +158,27 @@ defmodule Bamboo.SendgridAdapter do
 
   defp maybe_put_x_smtp_api(body, %Email{private: %{"x-smtpapi" => fields}} = email) do
     # SendGrid will error with empty bodies, even while using templates.
-    # Sets a default `text_body` if one is not specified, allowing the consumer
-    # to neglect doing so themselves.
-    case {email.text_body, email.html_body} do
-      {nil, nil} -> put_text_body(body, %Email{email | text_body: " "})
-      _ -> raise """
-      cannot set the body and use a SendGrid template at the same time. Please remove the body or SendGrid template
-      """
-    end |> Map.put("x-smtpapi", Poison.encode!(fields))
+    # Sets a default `text_body` and 'html_body' if either are not specified,
+    # allowing the consumer to neglect doing so themselves.
+    body = if is_nil(email.text_body) do
+      put_text_body(body, %Email{email | text_body: " "})
+    else
+      body
+    end
+
+    body = if is_nil(email.html_body) do
+      put_html_body(body, %Email{email | html_body: " "})
+    else
+      body
+    end
+
+    body = if is_nil(email.subject) do
+      put_subject(body, %Email{email | subject: " "})
+    else
+      body
+    end
+
+    Map.put(body, "x-smtpapi", Poison.encode!(fields))
   end
   defp maybe_put_x_smtp_api(body, _), do: body
 
--- a/lib/bamboo/adapters/sendgrid_adapter.ex
+++ b/lib/bamboo/adapters/sendgrid_adapter.ex
@@ -115,7 +115,7 @@ defmodule Bamboo.SendgridAdapter do
     |> put_subject(email)
     |> put_html_body(email)
     |> put_text_body(email)
-    |> put_x_smtp_api(email)
+    |> maybe_put_x_smtp_api(email)
   end
 
   defp put_from(body, %Email{from: {"", address}}), do: Map.put(body, :from, address)
@@ -156,16 +156,16 @@ defmodule Bamboo.SendgridAdapter do
   defp put_text_body(body, %Email{text_body: nil}), do: body
   defp put_text_body(body, %Email{text_body: text_body}), do: Map.put(body, :text, text_body)
 
-  defp put_x_smtp_api(body, %Email{private: %{"x-smtpapi" => fields}} = email) do
+  defp maybe_put_x_smtp_api(body, %Email{private: %{"x-smtpapi" => fields}} = email) do
     # SendGrid will error with empty bodies, even while using templates.
     # Sets a default `text_body` if one is not specified, allowing the consumer
-    # to negelect doing so themselves.
+    # to neglect doing so themselves.
     case {email.text_body, email.html_body} do
       {nil, nil} -> put_text_body(body, %Email{email | text_body: " "})
       _ -> body
     end |> Map.put("x-smtpapi", Poison.encode!(fields))
   end
-  defp put_x_smtp_api(body, _), do: body
+  defp maybe_put_x_smtp_api(body, _), do: body
 
   defp put_addresses(body, field, addresses), do: Map.put(body, field, addresses)
   defp put_names(body, field, names) do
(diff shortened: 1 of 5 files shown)


rafroehlich2:
Outstanding issue is getting the raw JSON `x-smtpapi` passed without escape characters. SendGrid does not appear to handle them well.

paulcsmith:
This looks pretty cool and I can see how it would be useful for people. Can you confirm that you've tried this out and it works? I have a few style comments and questions that I'll try to leave in the next few days :D

rafroehlich2:
I can confirm that the template functionality works. Looking forward to the feedback!

lib/bamboo/adapters/sendgrid_adapter.ex
```diff
@@ -107,6 +107,7 @@ defmodule Bamboo.SendgridAdapter do
     |> put_subject(email)
     |> put_html_body(email)
     |> put_text_body(email)
+    |> put_x_smtp_api(email)
```
paulcsmith: How about `maybe_put_x_smtp_api` so it's clear that it may or may not happen?

lib/bamboo/adapters/sendgrid_adapter.ex
```diff
@@ -147,6 +148,17 @@ defmodule Bamboo.SendgridAdapter do
   defp put_text_body(body, %Email{text_body: nil}), do: body
   defp put_text_body(body, %Email{text_body: text_body}), do: Map.put(body, :text, text_body)
 
+  defp put_x_smtp_api(body, %Email{private: %{"x-smtpapi" => fields}} = email) do
+    # SendGrid will error with empty bodies, even while using templates.
+    # Sets a default `text_body` if one is not specified, allowing the consumer
+    # to negelect doing so themselves.
```
paulcsmith: Usually I prefer no comments, but in this case I think this is a good idea!

There is one typo though. negelect -> neglect

lib/bamboo/adapters/sendgrid_adapter.ex
```diff
@@ -147,6 +148,17 @@ defmodule Bamboo.SendgridAdapter do
...
+    case {email.text_body, email.html_body} do
+      {nil, nil} -> put_text_body(body, %Email{email | text_body: " "})
+      _ -> body
+    end |> Map.put("x-smtpapi", Poison.encode!(fields))
+  end
```
paulcsmith: Is there any reason someone would set a body *and* use a Sendgrid template? If not then maybe we should raise if either of the bodies are set and then always set an empty text body. What do you think?

lib/bamboo/adapters/sendgrid_helper.ex
```diff
@@ -0,0 +1,77 @@
+defmodule Bamboo.SendgridHelper do
+  @moduledoc """
+  Functions for using features specific to Sendgrid template substitution tags.
```
paulcsmith: I think the moduledoc should be bit more generic. maybe `Functions for using feature specific to Sendgrid`?

lib/bamboo/adapters/sendgrid_helper.ex
```diff
@@ -0,0 +1,77 @@
...
+
+  Only one template can be specified, but multiple tags can be added to the
```
paulcsmith: This seems like something that might need a header. Like `## Using templates and tags together`

Also, should it be "tags" or should it be "substitutions"?

lib/bamboo/adapters/sendgrid_helper.ex
```diff
@@ -0,0 +1,77 @@
...
+  SendGrid. If called multiple times, the last template specified will be used.
+
+  ## Example
+
+      email
+      |> with_template("80509523-83de-42b6-a2bf-54b7513bd2aa")
+  """
+  def with_template(%Email{private: %{"x-smtpapi" => _}} = email, template_id) when byte_size(template_id) == @id_size do
+    fields = email.private["x-smtpapi"]
+             |> Map.merge(%{"filters" => build_template_filter(template_id)})
```
paulcsmith: The style we've been using is to only indent this by two spaces

lib/bamboo/adapters/sendgrid_helper.ex
```diff
@@ -0,0 +1,77 @@
...
+
+  alias Bamboo.Email
+
+  @id_size 36
+
+  @doc """
+  Specify the template for SendGrid to use for the substitutions
+
+  The `template_id` must be a valid `UUID` as generated for each template by
+  SendGrid. If called multiple times, the last template specified will be used.
```
paulcsmith: I'm not sure I understand this part. What is the use case for calling this multiple times, and why should it use the last template? Forgive the naivety. I haven't used Sendgrid templates so maybe this is super obvious and I just don't get it!

test/lib/bamboo/adapters/sendgrid_helper_test.exs
```diff
@@ -0,0 +1,76 @@
...
+    assert email.private["x-smtpapi"] == %{
+                                            "filters" => %{
+                                              "templates" => %{
+                                                "settings" => %{
+                                                  "enable" => 1,
+                                                  "template_id" => @template_id
+                                                }
+                                              }
+                                            }
+                                          }
```
paulcsmith: We typically keep this aligned to the left

test/lib/bamboo/adapters/sendgrid_helper_test.exs
```diff
@@ -0,0 +1,76 @@
...
+
+  import Bamboo.SendgridHelper
+
+  @template_id "80509523-83de-42b6-a2bf-54b7513bd2aa"
+
+  setup do
+    {:ok, email: Bamboo.Email.new_email}
+  end
+
+  test "(with_template) adds the correct template", %{email: email} do
```
paulcsmith: What do you think of `test "with_template/2`?

test/lib/bamboo/adapters/sendgrid_helper_test.exs
```diff
@@ -0,0 +1,76 @@
...
+                                                  "enable" => 1,
+                                                  "template_id" => @template_id
+                                                }
+                                              }
+                                            }
+                                          }
+  end
+
+  test "(with_template) raises on non-UUID `template_id`", %{email: email} do
+    assert_raise FunctionClauseError, fn ->
```
paulcsmith: Rather than raising a FunctionClauseError, what do you think of adding a catch all function that raises with a clear message?

Something like:

```
raise "expected the template ID to be a UUID that is 36 characters long, instead got: #{uuid}"
```

lib/bamboo/adapters/sendgrid_helper.ex
```diff
@@ -0,0 +1,77 @@
...
+
+    fields = email.private["x-smtpapi"]
+      |> Map.merge(%{"sub" => substitutions})
+
+    email |> Email.put_private("x-smtpapi", fields)
+  end
+
+  def substitute(email, tag, value) when is_binary(tag) do
+    email |> Email.put_private("x-smtpapi", %{"sub" => %{tag => [value]}})
+  end
```
paulcsmith: Rather than splitting these into two functions, could we do something like

```elixir
def substitute(email, tag, value) when is_binary(tag) do
  email
  |> Map.put_new("x-smtpapi", %{"sub" => %{}})
  |> ...do_everything_else
end
```

I think that might be more clear. What do you think? Something similar could also be done for `with_template` I think

test/lib/bamboo/adapters/sendgrid_adapter_test.exs
```diff
@@ -116,6 +116,22 @@ defmodule Bamboo.SendgridAdapterTest do
...
+    )
+
+    email
+    |> Bamboo.SendgridHelper.with_template("a4ca8ac9-3294-4eaf-8edc-335935192b8d")
+    |> Bamboo.SendgridHelper.substitute("%foo%", "bar")
+    |> SendgridAdapter.deliver(@config)
+
+    assert_receive {:fake_sendgrid, %{params: params}}
+    assert params["text"]
+    assert params["x-smtpapi"]
```
paulcsmith: Do you think it would be good to test that the values are also correct?

paulcsmith:
Hey @rafroehlich2 Thanks a ton for this. I left some comments on it. Let me know if you have any questions. I'm excited to get this in :D

lib/bamboo/adapters/sendgrid_adapter.ex
rafroehlich2: I do like the explicitness of raising when the developer has not properly used the API. Is there a general pattern for the exception that you would prefer?

lib/bamboo/adapters/sendgrid_helper.ex
rafroehlich2: That is probably a good idea. Are you thinking of just altering the first two sections or also abandon the provided example usage?

lib/bamboo/adapters/sendgrid_helper.ex
rafroehlich2: I was just attempting to err on the side of being explicit, that if someone is building up a message progressively across multiple functions, the last template specified would be used. This is probably obvious in intent and could be removed.

test/lib/bamboo/adapters/sendgrid_helper_test.exs
rafroehlich2: Do you have an opinion on using a simple `RuntimeError` or using a more explicit `SendgridTemplateError`? lt could serve the purpose both for this as well as the case of setting a body and a template at the same time.

lib/bamboo/adapters/sendgrid_adapter.ex
paulcsmith: I think raising a generic error would be fine. Something like

```
raise """
cannot set the body and use a SendGrid template at the same time. Please remove the body or the SendGrid template
"""
```

What do you think?

lib/bamboo/adapters/sendgrid_helper.ex
```diff
@@ -0,0 +1,82 @@
...
+  tags in the template.
+
+  ## Example
+
+      email
+      |> substitute("%name%", "Jon Snow")
+  """
+  def substitute(email, tag, value) do
+    case is_binary(tag) do
+      false ->
```
paulcsmith: Since this is a boolean I think an `if` might make more sense here

```elixir
if is_binary(tag) do
  # send it...
else
  # raise...
end
```

test/lib/bamboo/adapters/sendgrid_helper_test.exs
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  test "with_template/2 raises on non-UUID `template_id`", %{email: email} do
+    assert_raise RuntimeError, fn ->
+      email |> with_template("not a UUID")
+    end
+  end
+
+  test "with_template/2 uses the last specified template", %{email: email} do
+    last = "355d0197-ecf5-4268-aa8b-2c0502aec406"
```
paulcsmith: How about `last_template_id`?

paulcsmith:
I think this looks great! I (hopefully) answered your question and left a couple comments. I'll merge once the error is added. Thanks so much for this (and sorry for the late response. Been a bit busy this summer :D)

rafroehlich2:
No worries. The error in the tests are a result of the ExMachina issue #145. The individual tests for  SendgridAdapter and SendgridHelper all test green. I will change the dependency version once it is updated with the fix.

paulcsmith:
@rafroehlich2 I believe you can rebase against the latest on master and it should pass on CI! If it does I'll go ahead and merge this in. Thanks for your work on it 🎉

rafroehlich2:
One more fix coming down the pipe so hold off merging.

paulcsmith:
I also think this might need to be rebased because there are a bunch of commits in this PR

rafroehlich2:
This is ready to be merged. Let me know if you want this further rebased.

madshargreave:
Status on merge?

paulcsmith:
Thank you so much @rafroehlich2! I think this will be very helpful to SendGrid users

PR #164, opened 2016-05-23
Add type specs for address functions

drapergeek:
I recently ran into an issue where I was trying to figure out why a comma separated string wasn't working  for the `to` and I needed to see what type was supported when calling the `to` function. I think by adding these it makes it easier for users to see what options are passed but also it means that users could perform static analysis to make sure they're calling it correctly.

### DIFF
--- a/lib/bamboo/email.ex
+++ b/lib/bamboo/email.ex
@@ -68,11 +68,12 @@ defmodule Bamboo.Email do
   """
 
   @type address :: String.t | {String.t, String.t}
+  @type address_list :: nil | address | [address] | any
 
   @type t :: %__MODULE__{
-    to: nil | any,
-    cc: nil | any,
-    bcc: nil | any,
+    to: address_list,
+    cc: address_list,
+    bcc: address_list,
     subject: nil | String.t,
     html_body: nil | String.t,
     text_body: nil | String.t,
@@ -94,7 +95,8 @@ defmodule Bamboo.Email do
 
   alias Bamboo.Email
 
-  @attribute_pipe_functions ~w(from to cc bcc subject text_body html_body)a
+  @address_functions ~w(from to cc bcc)a
+  @attribute_pipe_functions ~w(subject text_body html_body)a
 
   @doc """
   Used to create a new email
@@ -111,6 +113,22 @@ defmodule Bamboo.Email do
   @spec new_email(Enum.t) :: __MODULE__.t
   def new_email(attrs \\ []) do
     struct!(%__MODULE__{}, attrs)
+  end
+
+  for function_name <- @address_functions do
+    @doc """
+    Sets the `#{function_name}` on the email.
+
+    You can pass in a string, list of strings,
+    or anything that implements the `Bamboo.Formatter` protocol.
+
+        new_email
+        |> #{function_name}(["sally@example.com", "james@example.com"])
+    """
+    @spec unquote(function_name)(__MODULE__.t, address_list) :: __MODULE__.t
+    def unquote(function_name)(email, attr) do
+      Map.put(email, unquote(function_name), attr)
+    end
   end
 
   for function_name <- @attribute_pipe_functions do


lib/bamboo/email.ex
```diff
@@ -68,11 +68,12 @@ defmodule Bamboo.Email do
   """
 
   @type address :: String.t | {String.t, String.t}
+  @type address_list :: nil | address | [address]
```
paulcsmith: or `any` because it can be anything that implements Bamboo.Formatter protocol

lib/bamboo/email.ex
```diff
@@ -113,6 +115,16 @@ defmodule Bamboo.Email do
     struct!(%__MODULE__{}, attrs)
   end
 
+  for function_name <- @address_functions do
+    @doc """
+    Sets the #{function_name} on the email
```
paulcsmith: Maybe add an example?

```elixir
    email |> #{function_name}("chase@gmail.com")
```

It would be good to run `mix docs` to make sure it looks ok though

paulcsmith:
This is a great idea! Just left a couple comments :)

lib/bamboo/email.ex
drapergeek: Can you explain that more?

I don't like `any` if it can be avoided. That doesn't allow type checking anymore.

paulcsmith: You can implement the Bamboo.Formatter protocol for anything. Maps, structs, atoms, boo leans, anything that protocols allow you to implement. So any value you want could be formatted. We could say that it accepts a map or list of maps, but that wouldn't really be technically correct :S

I suppose we could just say that anything other than a map/struct would be weird to format with a protocol and just add a type for that but I'm not totally sure about that. 

Or maybe there a way to add a type for anything that implements a protocol? That would work best if possible.

drapergeek:
@paulcsmith updated!

lib/bamboo/email.ex
```diff
@@ -117,7 +117,12 @@ defmodule Bamboo.Email do
 
   for function_name <- @address_functions do
     @doc """
-    Sets the #{function_name} on the email
+    Sets the `#{function_name}` on the email
+
+    ```
+    new_email
+    |> #{function_name}(["sally@example.com", "james@example.com"])
```
paulcsmith: We've been indenting by 4 spaces instead of using ```. I think after that's changes this can be merged in :D

paulcsmith: Also, what do you think of linking to the docs in Bamboo.Formatter? Something like

```
You can pass in a string, list of string, or anything that implements the `Bamboo.Formatter` protocol. See `Bamboo.Formatter` for more examples.
```

That might help people figure out what they can use for recipients. What do you think?

drapergeek: > We've been indenting by 4 spaces instead of using ```. I think after that's changes this can be merged in :D


Uh...you sure about that? https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/email.ex#L21-L34

https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/email.ex#L42-L67

paulcsmith: Almost everywhere else we are

https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapter.ex#L15
https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/formatter.ex#L39
https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/phoenix.ex#L12
https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/test.ex#L31
https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/plug/email_preview_plug.ex#L13
https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/mailgun_adapter.ex#L10

Thanks for pointing out Bamboo.Email though. Those should be changed to 4 spaces of indentation. Oops!

paulcsmith: Added a PR to change those examples: https://github.com/thoughtbot/bamboo/pull/165

paulcsmith:
LGTM! This will make it a lot easier for people to figure out what can be done with recipients. Thank you very much!

Issue #166, opened 2016-05-25
adapter config + conform doesnt work


smdern:
We use conform for setting sensitive application configuration e.g. Api Keys. Because the config is vetted during compilation any mailer configuration wouldn't be loaded yet by conform.

paulcsmith:
@smdern I'm not familiar with Conform so I'm not sure I understand the problem. Is the config not available at compile time when using Conform? Could you paste an example of what you prod.exs file looks like and what your mailer looks like so I could get a better understanding of the problem? Thanks!

smdern:
Conform takes a `.conf` and maps to the app's config during startup. 
https://github.com/bitwalker/conform

config/prod.exs
```elixir
config :my_app, MyApp.Mailer, adapter: Bamboo.MandrillAdapter
```
config/myapp.schema.exs
```elixir
[
  mappings: [
    "mandrill.api_key": [
       to: "my_app.Elixir.MyApp.Mailer.api_key",
       datatype: :binary,
    ]
  ]
]
```

smdern:
I have a branch that supports generating the config during compile time (default) and during runtime via a `dynamic_config` option https://github.com/thoughtbot/bamboo/pull/167

geofflane:
👍
Conform modifies the Elixir sys.config when the application is started which allows you to compile your application one time with Exrm and deploy it to multiple different environments, each with their own configurations. e.g. Staging and Production.

Without this you would have to compile a different version that would work in a staging environment

paulcsmith:
Thanks for the PR. This makes sense but I wonder if it would be best to choose just one way rather than having two ways of parsing the options. I think it might be better to just parse at runtime instead

What do you think about that?

geofflane:
@paulcsmith I agree with just parsing at runtime. Unless there was some significant performance cost (which I would doubt).

smdern:
Okay. I'll re-submit a PR w/ only runtime. I don't think any of the tests really cared about it being handled during runtime.

smdern:
https://github.com/thoughtbot/bamboo/pull/170

paulcsmith:
Closed by #170

PR #167, opened 2016-05-25
Add dynamic_config option

smdern:
When using https://github.com/bitwalker/conform the app config isn't loaded during compile time. This adds an option for the config to be generated during runtime vs compile time.

https://github.com/thoughtbot/bamboo/issues/166


paulcsmith:
Used #170 instead

PR #170, opened 2016-05-28
Build mailer config during runtime

smdern:
Fixes - https://github.com/thoughtbot/bamboo/issues/166

### DIFF
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -62,20 +62,21 @@ defmodule Bamboo.Mailer do
 
   defmacro __using__(opts) do
     quote bind_quoted: [opts: opts] do
-      config = Bamboo.Mailer.parse_opts(__MODULE__, opts)
-
-      @adapter config.adapter
-      @config config
-
       @spec deliver_now(Bamboo.Email.t) :: Bamboo.Email.t
       def deliver_now(email) do
-        Bamboo.Mailer.deliver_now(@adapter, email, @config)
+        config = get_config
+        Bamboo.Mailer.deliver_now(config.adapter, email, config)
       end
 
       @spec deliver_later(Bamboo.Email.t) :: Bamboo.Email.t
       def deliver_later(email) do
-        Bamboo.Mailer.deliver_later(@adapter, email, @config)
+        config = get_config
+        Bamboo.Mailer.deliver_later(config.adapter, email, config)
       end
+
+      otp_app = Keyword.fetch!(opts, :otp_app)
+
+      defp get_config, do: Bamboo.Mailer.parse_opts(__MODULE__, unquote(otp_app))
 
       def deliver(_email) do
         raise """
@@ -184,9 +185,8 @@ defmodule Bamboo.Mailer do
   end
 
   @doc false
-  def parse_opts(mailer, opts) do
-    otp_app = Keyword.fetch!(opts, :otp_app)
-    config = Application.fetch_env!(otp_app, mailer) |> Enum.into(%{})
+  def parse_opts(mailer, otp_app) do
+    config = Application.fetch_env!(otp_app, mailer) |> Map.new
 
     config.adapter.handle_config(config)
     |> Map.put_new(:deliver_later_strategy, Bamboo.TaskSupervisorStrategy)
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -62,21 +62,22 @@ defmodule Bamboo.Mailer do
 
   defmacro __using__(opts) do
     quote bind_quoted: [opts: opts] do
+
       @spec deliver_now(Bamboo.Email.t) :: Bamboo.Email.t
       def deliver_now(email) do
-        config = get_config
+        config = build_config
         Bamboo.Mailer.deliver_now(config.adapter, email, config)
       end
 
       @spec deliver_later(Bamboo.Email.t) :: Bamboo.Email.t
       def deliver_later(email) do
-        config = get_config
+        config = build_config
         Bamboo.Mailer.deliver_later(config.adapter, email, config)
       end
 
       otp_app = Keyword.fetch!(opts, :otp_app)
 
-      defp get_config, do: Bamboo.Mailer.parse_opts(__MODULE__, unquote(otp_app))
+      defp build_config, do: Bamboo.Mailer.build_config(__MODULE__, unquote(otp_app))
 
       def deliver(_email) do
         raise """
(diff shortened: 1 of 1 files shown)


lib/bamboo/mailer.ex
```diff
@@ -64,17 +64,14 @@ defmodule Bamboo.Mailer do
     quote bind_quoted: [opts: opts] do
       config = Bamboo.Mailer.parse_opts(__MODULE__, opts)
```
geofflane: Is this line still needed? Should it always use the function instead?

smdern: Thanks. good catch.

smdern: `parse_opts` was doing a `Keyword.fetch!(opts, :otp_app)` to enforce that it wasn't omitted. Going to include that in the macro so that can get caught during compile time.

lib/bamboo/mailer.ex
```diff
@@ -62,19 +62,14 @@ defmodule Bamboo.Mailer do
...
     quote bind_quoted: [opts: opts] do
-      config = Bamboo.Mailer.parse_opts(__MODULE__, opts)
-
-      @adapter config.adapter
-      @config config
-
       @spec deliver_now(Bamboo.Email.t) :: Bamboo.Email.t
       def deliver_now(email) do
-        Bamboo.Mailer.deliver_now(@adapter, email, @config)
+        Bamboo.Mailer.deliver_now(adapter, email, config)
```
geofflane: Random thoughts: This parses the config 2x per delivery, once to get the config, but again to get adapter which just calls `config.adapter`. Maybe something like:

```
conf = config
Bamboo.Mailer.deliver_now(conf.adapter, email, conf)
```

Premature optimization. :)

smdern: :+1: Appreciate the feedback.

lib/bamboo/mailer.ex
```diff
@@ -184,8 +184,7 @@ defmodule Bamboo.Mailer do
   end
 
   @doc false
-  def parse_opts(mailer, opts) do
-    otp_app = Keyword.fetch!(opts, :otp_app)
+  def parse_opts(mailer, otp_app) do
```
smdern: ❓ This makes it so that the omitting `:otp_app` in a `Mailer` is caught during compile time, but it changes a public method signature. Can add an additional clause `def parse_opts(mailer, opts, otp_app)` if concerned

geofflane:
👍

lib/bamboo/mailer.ex
```diff
@@ -184,8 +184,7 @@ defmodule Bamboo.Mailer do
...
     config = Application.fetch_env!(otp_app, mailer) |> Enum.into(%{})
```
paulcsmith: Since you're working on the config, could you change `Enum.into(%{})` into `Map.new`? That is faster and I think a bit more clear.

lib/bamboo/mailer.ex
```diff
@@ -62,19 +62,16 @@ defmodule Bamboo.Mailer do
...
-        Bamboo.Mailer.deliver_now(@adapter, email, @config)
+        conf = config
+        Bamboo.Mailer.deliver_now(conf.adapter, email, conf)
       end
 
       @spec deliver_later(Bamboo.Email.t) :: Bamboo.Email.t
       def deliver_later(email) do
-        Bamboo.Mailer.deliver_later(@adapter, email, @config)
+        conf = config
+        Bamboo.Mailer.deliver_later(conf.adapter, email, conf)
```
paulcsmith: Rather than creating an intermediate variable `conf`, you could use `config` in the first and third argument. I think that might work. It might be *slightly* slower, but that is probably ok. Alternatively I would assign to `config` and rename the function to `get_config`

lib/bamboo/mailer.ex
```diff
@@ -84,6 +81,9 @@ defmodule Bamboo.Mailer do
         Use deliver_now/1 to send right away, or deliver_later/1 to send in the background.
         """
       end
+
+      otp_app = Keyword.fetch!(opts, :otp_app)
+      defp config, do: Bamboo.Mailer.parse_opts(__MODULE__, unquote(otp_app))
```
paulcsmith: Could this be moved below `deliver_later` so it's closer to the function that uses it?

lib/bamboo/mailer.ex
```diff
@@ -84,6 +81,9 @@ defmodule Bamboo.Mailer do
         Use deliver_now/1 to send right away, or deliver_later/1 to send in the background.
         """
       end
+
+      otp_app = Keyword.fetch!(opts, :otp_app)
```
paulcsmith: I think a newline between this and `defp config` would be nice. What do you think?

lib/bamboo/mailer.ex
paulcsmith: I think it's ok because it's not documented and probably shouldn't be used by other devs anyway :)

paulcsmith:
Left a few small comments. Overall this looks great though :) Can you ping me when you've had a chance to implement those suggestions? Then I'll merge this in. Thanks!

lib/bamboo/mailer.ex
smdern: going w/ the `get_config` and `config` approach. I'm :+1: for readability

smdern:
> Left a few small comments. Overall this looks great though :) Can you ping me when you've had a chance to implement those suggestions? Then I'll merge this in. Thanks!

@paulcsmith feedback addressed.

lib/bamboo/mailer.ex
```diff
@@ -62,21 +62,22 @@ defmodule Bamboo.Mailer do
...
       @spec deliver_later(Bamboo.Email.t) :: Bamboo.Email.t
       def deliver_later(email) do
-        Bamboo.Mailer.deliver_later(@adapter, email, @config)
+        config = get_config
+        Bamboo.Mailer.deliver_later(config.adapter, email, config)
       end
 
+      otp_app = Keyword.fetch!(opts, :otp_app)
+
+      defp get_config, do: Bamboo.Mailer.parse_opts(__MODULE__, unquote(otp_app))
```
smdern: I'm thinking this should be called `build_config` so it implies that there is some sort of work that has to be done whenever it's called

paulcsmith: Yeah I like that!

smdern: Actually how do you feel about `Bamboo.Mailer.build_config(mailer, otp_app)` and then we can just inline it in the functions as `config = Bamboo.Mailer.build_config(mailer, unquote(otp_app)` ? Seems like `build_config` would be a better public function than `parse_opts`.

paulcsmith: I think I'd have to see it first, but it sounds pretty good. Definitely worth a try

smdern: @paulcsmith d226f86

smdern:
@paulcsmith Made the changes.

paulcsmith:
Thank you @smdern!

PR #172, opened 2016-06-01
move preview of HTML email to a separate view

jerodsanto:
When previewing HTML emails that include their own styles, the current implementation will allow said styles to "leak" out and disrupt the preview index styles. Example:

2016-06-01 at 10 49 am

To plug the leak, I've moved the HTML previews to their own view altogether. Example:

2016-06-01 at 11 22 am

Other possible ways to resolve this:

1. Shadow DOM – I went down this route and got it working, only to find that you cannot have a `<link>` element in a shadow tree. This kills anyone using external styles (and letting their delivery provider inline them).
2. iframe – I think this might work if we set the html body to the `srcdoc` attribute, but my first pass failed and then we'd have to resize the iframes to fit the content. Seems kludgy.

I'm open to whatever works best. Just holler and I'l make adjustments as necessary. 🍻

### DIFF
--- a/lib/bamboo/plug/email_preview_plug.ex
+++ b/lib/bamboo/plug/email_preview_plug.ex
@@ -50,6 +50,14 @@ defmodule Bamboo.EmailPreviewPlug do
     end
   end
 
+  get "/:id/html" do
+    if email = SentEmail.get(id) do
+      conn |> send_resp(:ok, email.html_body)
+    else
+      conn |> render(:not_found, "email_not_found.html")
+    end
+  end
+
   defp all_emails do
     SentEmail.all
   end
--- a/lib/bamboo/plug/index.html.eex
+++ b/lib/bamboo/plug/index.html.eex
@@ -159,7 +159,11 @@
         <section class="email-preview-bodies-container">
           <h3 class="email-preview-body-label">HTML body</h3>
           <p class="email-preview-body">
-            <%= @selected_email.html_body %>
+            <%= if is_nil(@selected_email.html_body) do %>
+              None
+            <% else %>
+              <a href="<%= "#{@base_path}/#{Bamboo.SentEmail.get_id(@selected_email)}/html" %>">Click Here to View</a>
+            <% end %>
           </p>
 
           <h3 class="email-preview-body-label">Text Body</h3>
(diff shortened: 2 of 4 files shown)


paulcsmith:
I think having the styles show in the preview is pretty necessary, but I agree that it should only apply to the section of the HTML that the email is in. I think an iframe makes the most sense, but maybe Shadow DOM works as you suggested. I've never used it though so I'm unsure!

I'd be happy to take a look at using an iframe/shadow DOM sometime. I think an iframe should work though. Do you have the code you tried that didn't work?

jerodsanto:
Shadow DOM doesn't work for the reason I said above: browsers won't load `<link>` elements inside the shadow tree, so any externally referenced stylesheets won't load.

The iframe attempts I made were:

```eex
<p class="email-preview-body">
  <iframe srcdoc="<%= @selected_email.html_body %>"></iframe>
</p>
```

and then trying to escape double quotes:

```eex
<p class="email-preview-body">
  <iframe srcdoc="<%= ~s(#{@selected_email.html_body}) %>"></iframe>
</p>
```

Neither of those outputted a valid attribute string for it to render. Also worth noting, `srcdoc` is not supported in *any* version of IE/Edge. Not sure how important that is to you.

paulcsmith:
Oh sorry, should have read the note about Shadow DOM more closely. Maybe using `Plug.HTML.html_escape` would work?

jerodsanto:
ACTUALLY combining both methods (a separate view to render the html + an iframe with said view as its `src`) is probably the _best_ solution:

2016-06-01 at 1 02 pm

Now we just need a little JS to resize the iframe to fit its content (or we can set it to a static width/height and let the email scroll) and it should be 👍

paulcsmith:
That’s a good idea. Loading the view in the iframe sounds like a great idea. I think that we could set the height and width with CSS. I can take a look at that once you’ve got the iframe loading the html view :D

jerodsanto:
@paulcsmith done. I set the iframe width to 100% and added a js function to adjust the height to fit the iframe content. Looks like this (my only test case so far):

2016-06-01 at 2 02 pm

Let me know what you think.

paulcsmith:
Can you add `border: none;` to get rid of the border around the iframe? Other than that this looks awesome. I'll have to do a more thorough review on Friday, but after a quick pass it looks awesome :)

jerodsanto:
Done. Feel free to improve/cleanup the CSS and JS bits. I'm not tied to any of that.

paulcsmith:
Just tried this out. I found one small bug, but I'll open a separate issue for it! Thank you so much

Issue #110, opened 2016-04-01
Add example of what a layout file would look like


paulcsmith:
Add an example of what a layout file would look like when used with `Bamboo.Phoenix`. I always forget how to call `render` with the assigns and I'm sure others will have to look it up as well. Might be worth including an example to save some time and make it easier to use the library.

jerodsanto:
This would be super useful. Specifically, it'd be nice to see how we can use route helpers in layouts/templates since `@conn` isn't a thing.

_(I have a decent/simple sign in email w/ layout example I can provide once it's 💯 . Just let me know if/where you want it.)_

paulcsmith:
@jerodsanto It would be wonderful if you could paste an example here or open up a PR. Especially if it includes using `MyApp.Endpoint` in URL helpers so that people can see how to use those. Thanks for offering to help!

jerodsanto:
@paulcsmith happy to PR. Where do you want it the example to go? `bamboo/lib/bamboo/phoenix.ex`?

There are a few files...

paulcsmith:
Yes, Bamboo.Phoenix would be perfect!

paulcsmith:
Closed by #173

PR #173, opened 2016-06-02
add Full HTML Layout Example to Bamboo.Phoenix docs

jerodsanto:
As discussed in #110. Feedback welcome. 🍻

### DIFF
--- a/lib/bamboo/phoenix.ex
+++ b/lib/bamboo/phoenix.ex
@@ -51,6 +51,49 @@ defmodule Bamboo.Phoenix do
           |> render("text_email.text")
         end
       end
+
+  ## Full HTML Layout Example
+
+      # web/email.ex
+      defmodule Changelog.Email do
+        use Bamboo.Phoenix, view: Changelog.EmailView
+
+        def sign_in_email(person) do
+          base_email
+          |> to(person)
+          |> subject("Your Sign In Link")
+          |> assign(:person, person)
+          |> render(:sign_in)
+        end
+
+        defp base_email do
+          new_email
+          |> from("Rob Ot<robot@changelog.com>")
+          |> put_header("Reply-To", "editors@changelog.com")
+          |> put_html_layout({Changelog.LayoutView, "email.html"})
+        end
+      end
+
+      # web/views/email_view.ex
+      defmodule Changelog.EmailView do
+        use Changelog.Web, :view
+      end
+
+      # web/templates/layout/email.html.eex
+      <html>
+        <head>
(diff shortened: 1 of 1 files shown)


paulcsmith:
@jerodsanto This looks really good! I have a couple suggestions, but won't have much time until Friday. I'll leave some comments then. Thank you!

lib/bamboo/phoenix.ex
```diff
@@ -51,6 +51,49 @@ defmodule Bamboo.Phoenix do
           |> render("text_email.text")
         end
       end
+
+  ## Full HTML Layout Example
+
+      # web/email.ex
+      defmodule Changelog.Email do
```
paulcsmith: Could you change all instance of `Changelog` to `MyApp`? That's the default namespace used in the Bamboo docs

lib/bamboo/phoenix.ex
```diff
@@ -51,6 +51,49 @@ defmodule Bamboo.Phoenix do
...
+          |> subject("Your Sign In Link")
+          |> assign(:person, person)
+          |> render(:sign_in)
+        end
+
+        defp base_email do
+          new_email
+          |> from("Rob Ot<robot@changelog.com>")
+          |> put_header("Reply-To", "editors@changelog.com")
+          |> put_html_layout({Changelog.LayoutView, "email.html"})
```
paulcsmith: Maybe add a code comment above this that says something like

```elixir
# This will use the "email.html.eex" file as a layout when rendering html emails. 
# Plain text emails will not use a layout unless you use `put_text_layout`
```

lib/bamboo/phoenix.ex
```diff
@@ -51,6 +51,49 @@ defmodule Bamboo.Phoenix do
...
+        <body>
+          <%= render @view_module, @view_template, assigns %>
+        </body>
+      </html>
+
+      # web/templates/email/sign_in.html.eex
+      <p><%= link "Sign In", to: sign_in_url(Changelog.Endpoint, :create, @person) %></p>
+
+      # web/templates/email/sign_in.text.eex
+      Sign In: <%= sign_in_url(Changelog.Endpoint, :create, @person) %>
```
paulcsmith: And maybe a comment here that says `# This will not be rendered within a layout because `put_text_layout` was not used`. What do you think?

paulcsmith:
@jerodsanto This looks awesome. Left just a couple comments and then this is ready to go!

jerodsanto:
@paulcsmith  cool, cool. Changes added. 💥

paulcsmith:
Thank you @jerodsanto!

jerodsanto:
🎉

PR #185, opened 2016-06-24
Remove warnings-as-errors until deps are updated

paulcsmith:
Right now CI isn't passing due to some deps that have warnings under the new version of Erlang. So I'll remove this and add it back in once we've had a chance to figure get it passing without any warnings again. For now this is making it harder for contributors to get things in so I'll remove it.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,4 +11,3 @@ dependencies:
 test:
   override:
     - mix test
-    - MIX_ENV=prod mix compile --warnings-as-errors


john-griffin:
Looks like regular `mix test` is throwing the same warnings.

paulcsmith:
Yeah the warnings shouldn't make the build fail though :S I'll have to dig into this more in a bit :)

paulcsmith:
Closed in favor of #186 ExMachina 1.0 works under Erlang OTP 19 now

PR #189, opened 2016-06-26
bump dependencies

princemaple:
Should run `mix deps.update --all` from time to time...

I was going to bring in the latest version of httpoison to get rid of the warnings that elixir 1.3 concerns about, but thought there could be other things that need upgrades. And indeed there are.


princemaple:
maybe you don't want the change in `mix.exs`, let me know.

paulcsmith:
@princemaple Thanks for running this. You're right that deps should be updated from time to time and I appreciate the PR!

PR #192, opened 2016-07-02
Refactor base uri

princemaple:
- Move base uri resolution to compile time
- Refactor and fix tests
- Extract the common part (Fake...) into a separate module

I've made sure the adapter tests are not coupled with the fake endpoint in https://github.com/thoughtbot/bamboo/pull/192/commits/2ae71d351bb2bee78a99cf16356749caa7839c04 with a tiny bit of meta. So later on when you want to extract the adapters out of the main project, you can simply move the adapter tests out with their respective source code.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,5 +1,3 @@
 use Mix.Config
 
-config :logger, level: :warn
-
 import_config "#{Mix.env}.exs"
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,30 +1,5 @@
-# This file is responsible for configuring your application
-# and its dependencies with the aid of the Mix.Config module.
 use Mix.Config
 
-# This configuration is loaded before any dependency and is restricted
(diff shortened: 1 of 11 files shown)


geofflane:
I think storing config values as module attributes breaks some deployment scenarios because the value is determined at compile time instead of at runtime from the configuration. e.g. If you managed your sys.config externally for different environments. This would also conflict with previous changes made to support using Conform for configuration management.

princemaple:
> because the value is determined at compile time instead of at runtime from the configuration.

Yeah I'm aware of that. It wasn't a problem for me, because I deploy everything with docker.

I'm OK with closing this one. Thanks for the review.

paulcsmith:
Thanks @princemaple for the PR and @geofflane for the review. I like the idea of extracting the fake server endpoints, but agree that it would suck to conflict with people's config management tools. Maybe there's a way to do that in the future

PR #195, opened 2016-07-07
Soft deprecates SendgridAdapter and adds SendGridAdapter

### LINKED ISSUE #141: Rename SendgridAdapter to SendGridAdapter
As discussed here: https://github.com/thoughtbot/bamboo/pull/140#discussion_r61640713

We should still have the old misspelled one, but `deliver_now` and `deliver_later` should raise and let people know that it has been renamed. This would be removed once Bamboo is at 1.0

jbernardo95:
This PR aims to fix #141

This PR includes:
- SendgridAdapter soft deprecation
- Adds SendGridAdapter and SendGridAdapterTest

Ps: This is my first Elixir PR feel free to tell me what's wrong 🙂


paulcsmith:
@jbernardo95 Congrats on your first Elixir PR! I'm a bit busy this week, but I will take a look at this next week if I can. If I don't leave a comment feel free to ping me because sometimes these slip my mind! Thanks!

jbernardo95:
@paulcsmith No problem, I will ping you here.

jbernardo95:
@paulcsmith Ping

paulcsmith:
@jbernardo95 Thanks for the ping! After thinking about this a bit I think there may be a way to handle this that makes it really clear to people what they need to do. I think we can basically rename `SendgridAdapter` to `SendGridAdapter` and then create another module called `SendgridAdapter` that implements the `deliver` function of `SendGridAdapter`, but instead of sending the email it raises an error letting the user know they need to rename the adapter. Something like this

```elixir
defmodule SendgridAdapter do
  @moduledoc false

  def deliver(email, config) do
    raise """
    Bamboo.SendgridAdapter has been renamed to SendGridAdapter (note the capital "G")

    Please use Bamboo.SendGridAdapter (with a capital "G") in your config. This was changed to       
    correctly match how SendGrid is spelled.
    """
  end
end
```

I don't think this needs to be tested since it is so simple and will be removed once Bamboo hits 1.0

Does that make sense?

jbernardo95:
@paulcsmith Completely agree 👍

I renamed the old `sendgrid_adapter.ex` to `send_grid_adapter.ex` including module name and where it is used.

And added the error raiding to the old `SendgridAdapter.deliver/2`.

lib/bamboo/adapters/send_grid_adapter.ex
```diff
@@ -0,0 +1,167 @@
...
+  Sends email using SendGrid's JSON API.
+
+  Use this adapter to send emails through SendGrid's API. Requires that an API
+  key is set in the config.
+
+  ## Example config
+
+      # In config/config.exs, or config.prod.exs, etc.
+      config :my_app, MyApp.Mailer,
+        adapter: Bamboo.SendgridAdapter,
```
geofflane: Doc comment should be updated to SendGridAdapter here as well.

lib/bamboo/adapters/sendgrid_adapter.ex
```diff
@@ -1,167 +1,12 @@
...
-        DATABASE_URL
-        SENDGRID_API_KEY
-      )
-      """
-      %ApiError{message: message}
-    end
-  end
+  @moduledoc false
 
   def deliver(email, config) do
```
geofflane: Since these are unused, they should be renamed to `_email` and `_config` so that you don't get build warnings.

geofflane:
This might be a problem on Mac OS X? Its filesystem is case-aware but not case sensitive. The generated Beam files both are called the "same" thing in that scenario. If I run the tests more than once in a row I get an error:

```
22:15:44.150 [error] beam/beam_load.c(1278): Error loading module 'Elixir.Bamboo.SendgridAdapter':
  module name in object code is Elixir.Bamboo.SendGridAdapter
```

jbernardo95:
@geofflane Both my fails, now fixed.

@paulcsmith @geofflane Any idea how to fix the `Error loading module 'Elixir.Bamboo.SendgridAdapter'` error ? Isn't this a beam loading issue ?

geofflane:
@jbernardo95 I think it's because the code now has `Bamboo.SendgridAdapter` and `Bamboo.SendGridAdapter` modules that have the same names and only vary by case. I think this might be an unsupported scenario?

jbernardo95:
@geofflane Indeed, but this sounds like a basic thing, shouldn't the build system support this ?

@paulcsmith Since this a temporary thing, should we worry about this ?

paulcsmith:
It sounds like the load error is a blocker for merging this.

I think the best solution is to remove `SendgridAdapter` completely and I'll wait to merge this until Bamboo 1.0.

Thoughts?

geofflane:
I wouldn't be opposed to merging it earlier and documenting it as a breaking change in the Changelog. I think most of us expect some breaking changes pre-1.0. Especially one that's just a single character change, that's pretty easy to manage.

jbernardo95:
I removed the old `SendgridAdapter` module, and the error is gone.

@paulcsmith I am with @geofflane, I think it is a pretty simple change, I see no problem seeing it merged on a `pre-1.0 release`. Possibly `0.7` ?

What do you think ?

(I will update the changelog according to your answer)

paulcsmith:
I think it is an easy to manage change, but it's not an easily discoverable change. Meaning that if I run `mix deps.update bamboo` and use Sendgrid, all of a sudden Sendgrid totally breaks. This would be super strange to me. I might expect an error, but not that the module no longer exists. I don't mind adding breaking changes like this if they are easily discoverable or are much better, but since this is a small change and one that people might get with a simple `mix deps.update` I'd rather not break things.

If we wait for `1.0` then people will need to *explicitly* update their mix file to `~> 1.0`. Since it's a big bump I don't feel as bad changing the behavior since you have to know about 1.0 and change it yourself. Does that make sense?

jbernardo95:
@paulcsmith Fair enough ! We should wait 🙂

paulcsmith:
Just a heads up. Been busy with some other projects, but I haven't forgotten about this. I think we're getting fairly close to 1.0. Once it does we'll get this in!

jbernardo95:
@paulcsmith Sure, no problem 😉

paulcsmith:
I rebased this and merged it here: https://github.com/thoughtbot/bamboo/commit/6b582f80781f0072bd4051084a3286991bfde2d0

Thanks @jbernardo95!

jbernardo95:
👍

PR #199, opened 2016-07-24
Fixed typo in test helper documentation

aaronrenner:
One character fix, but it makes this part of the documentation make more sense. 😄


paulcsmith:
Thank you! Good catch :)

PR #201, opened 2016-07-25
Drop dependency on httpoison, use hackney directly

michalmuskala:
This removes one dependency from the project, and I don't think it complicates it in any significant way.


paulcsmith:
I looked at this quickly on my phone and it seems good to me! I'll have to take a closer look on Friday when I have a bit more time but I think it'll be good to merge as-is. Thanks!

lib/bamboo/adapters/mailgun_adapter.ex
```diff
@@ -123,7 +130,7 @@ defmodule Bamboo.MailgunAdapter do
   defp put_text_body(body, %Email{text_body: text_body}), do: Map.put(body, :text, text_body)
 
   defp full_uri(config) do
-    (Application.get_env(:bamboo, :mailgun_base_uri) || @base_uri)
-    <> config.domain <> "/messages"
+    [Application.get_env(:bamboo, :mailgun_base_uri) || @base_uri,
```
paulcsmith: It seems strange that this return a list and not a string. Is that just a Hackney thing?

paulcsmith:
Just one question about using a list as the URI. Other than that, this looks great!

lib/bamboo/adapters/mailgun_adapter.ex
michalmuskala: hackney acceps iolists as url, I can revert it to append.

paulcsmith: I think I'd prefer that if it's cool with you :)

michalmuskala:
@paulcsmith updated

paulcsmith:
Thank you! :D Yay for one less dependency

PR #202, opened 2016-07-27
Editorial: Modified explanation on Phoenix vs Plug Router

davidkuhta:
Split explanation on implementation of the Bamboo.EmailPreviewPlug in a Phoenix Router vs a Plug Router.

### DIFF
--- a/lib/bamboo/plug/email_preview_plug.ex
+++ b/lib/bamboo/plug/email_preview_plug.ex
@@ -10,19 +10,24 @@ defmodule Bamboo.EmailPreviewPlug do
 
   ## Using with Plug or Phoenix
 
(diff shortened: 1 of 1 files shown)


lib/bamboo/plug/email_preview_plug.ex
```diff
@@ -10,19 +10,24 @@ defmodule Bamboo.EmailPreviewPlug do
 
   ## Using with Plug or Phoenix
 
-      # Make sure you are using Bamboo.LocalAdapter in your config
+      # Ensure you are using Bamboo.LocalAdapter in your config
```
davidkuhta: Wasn't sure that this should explicitly state "in your _dev_ config" give the references to `if Mix.env == :dev do` below?

PR #203, opened 2016-07-27
Fixes #174 - Set "text/html" Response Content Type for EmailPreviewPlug

### LINKED ISSUE #174: Bamboo.EmailPreviewPlug doesn't pass content-type (it renders as text in Chrome)
I'm on Linux Mint 17.2.

Using FireFox(45.0), the Bamboo.EmailPreviewPlug looks like this:

screenshot from 2016-06-04 17 50 09

However, in chrome (Chromium  49.0.2623.108)  it looks like this:

screenshot from 2016-06-04 17 50 41

This is probably caused by not passing a content-type in the return value, which makes Chrome fallback to 'text' behaviour.

davidkuhta:
Change: Added Pipe for |> Plug.Conn.put_resp_content_type("text/html")
Rationale: When getting "/:id/html" set the response content type to "text/html"

Fixes issue where the iframe in `index.html.eex` interprets the source as 'text' and embeds in a set of`<pre></pre>` tags, such that HTML tags aren't rendered.

Issued occurred for me on:
Chrome - Version 51.0.2704.103 (64-bit)
Safari - Version 9.1.2 (11601.7.6)

### DIFF
--- a/lib/bamboo/plug/email_preview_plug.ex
+++ b/lib/bamboo/plug/email_preview_plug.ex
@@ -52,7 +52,9 @@ defmodule Bamboo.EmailPreviewPlug do
 
   get "/:id/html" do
     if email = SentEmail.get(id) do
-      conn |> send_resp(:ok, email.html_body || "")
+      conn
+      |> Plug.Conn.put_resp_content_type("text/html")
+      |> send_resp(:ok, email.html_body || "")
     else
       conn |> render(:not_found, "email_not_found.html")
     end


lib/bamboo/plug/email_preview_plug.ex
```diff
@@ -52,7 +52,9 @@ defmodule Bamboo.EmailPreviewPlug do
 
   get "/:id/html" do
     if email = SentEmail.get(id) do
-      conn |> send_resp(:ok, email.html_body || "")
+      conn
+      |> Plug.Conn.put_resp_content_type("text/html")
```
paulcsmith: Could you also add this to line 59 so it works for `email_not_found`?

paulcsmith:
Thank you so much! I'm excited to get this in :D

paulcsmith:
Thanks @davidkuhta!

PR #210, opened 2016-09-02
Add PostmarkAdapter to README and description

pablo-co:
I created an adapter for [Postmark](https://postmarkapp.com/) which I've been using on production for a couple of weeks without a problem.

I decided to extract the adapter into it's own project so that others using Postmark can benefit from API only features such as templates. This change adds a reference to the adapter in the README and in the project's description.


paulcsmith:
Nice work, and thanks for the PR! Having extra adapters is always helpful

Issue #212, opened 2016-09-10
Email IDs in local adapter are case-sensitive


endersstocker:
This makes the Email Preview unusable behind an endpoint with lowercase path rewrite (a common practice).

paulcsmith:
Closed by #213

PR #213, opened 2016-09-10
Use case-insensitive sent email ID lookup

endersstocker:
This addresses #212, where the Email Preview was rendered unusable behind an endpoint with lowercase path rewrite (a common practice).

### DIFF
--- a/lib/bamboo/sent_email.ex
+++ b/lib/bamboo/sent_email.ex
@@ -89,7 +89,7 @@ defmodule Bamboo.SentEmail do
 
   defp do_get(id) do
     Enum.find all, nil, fn(email) ->
-      get_id(email) == id
+      email |> get_id |> String.downcase == String.downcase(id)
     end
   end
 
--- a/test/lib/bamboo/sent_email_test.exs
+++ b/test/lib/bamboo/sent_email_test.exs
@@ -37,6 +37,15 @@ defmodule Bamboo.SentEmailTest do
     assert %Bamboo.Email{subject: "Something"} = email
   end
 
+  test "get is case-insensitive" do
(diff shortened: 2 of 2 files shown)


lib/bamboo/sent_email.ex
```diff
@@ -89,7 +89,7 @@ defmodule Bamboo.SentEmail do
 
   defp do_get(id) do
     Enum.find all, nil, fn(email) ->
-      get_id(email) == id
+      get_id(email) == String.downcase(id)
```
paulcsmith: I'm not sure this needs to be downcast, since the URL is generated using the email id, and it should already be downcased at that point

paulcsmith:
Just one small comment. Thanks for the PR :)

lib/bamboo/sent_email.ex
endersstocker: While that's true for a lowercase rewrite, it’s not case-insensitive without this line. It would fail for an uppercase rewrite.

paulcsmith: I'm not sure how common that is but I think that's ok to keep it then. I think it would be good to add a test case where the passed in ID param is uppercase. That way we can ensure it doesn't break in the future

paulcsmith: If you need any help with this LMK and I'd be happy to give some pointers

endersstocker:
@paulcsmith I added a test and revised the implementation to make `SentEmail.get/1` truly case-insensitive. The randomly generated IDs themselves can remain mixed case.

paulcsmith:
Looks great. Thanks!

PR #215, opened 2016-09-20
Update README.md to working example

MatthewRDodds:
The `put_html_layout` and `put_text_layout` defined in the `Bamboo.Phoenix` [module](https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/phoenix.ex#L149-L195)


paulcsmith:
Good catch! Thank you @MatthewRDodds

PR #216, opened 2016-09-23
Update README.md to specify the domain configuration

OscarSwanros:
I couldn't find anywhere that said that for Mailgun it was necessary to include the domain on the Bamboo configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,9 @@ is done to make testing easier and to make emails easy to pipe/compose.
 # In your config/config.exs file
 config :my_app, MyApp.Mailer,
   adapter: Bamboo.MandrillAdapter,
-  api_key: "my_api_key"
-
+  api_key: "my_api_key",
+  domain: "my-domain.com" # Set the domain in case you're using Mailgun
+  
 # Somewhere in your application
 defmodule MyApp.Mailer do
   use Bamboo.Mailer, otp_app: :my_app


README.md
```diff
@@ -87,8 +87,9 @@ is done to make testing easier and to make emails easy to pipe/compose.
 # In your config/config.exs file
 config :my_app, MyApp.Mailer,
   adapter: Bamboo.MandrillAdapter,
-  api_key: "my_api_key"
-
+  api_key: "my_api_key",
+  domain: "my-domain.com" # Set the domain in case you're using Mailgun
```
paulcsmith: I think this is too specific to just Mailgun. I think instead there should be a comment that says "checks the docs for any adapter specific configuration" or something like that?

paulcsmith:
Hi @OscarSwanros. The docs for the MailgunAdapter are found here: https://hexdocs.pm/bamboo/0.7.0/Bamboo.MailgunAdapter.html. I can see how it might be confusing if someone doesn't know to look there though. I think a generic comment might be better.

What do you think?

README.md
paulcsmith: @OscarSwanros What do you think ^

optikfluffel:
Maybe also adding links to the corresponding docs for all the adapters would help too, since we list them anyway.

paulcsmith:
Closed by https://github.com/thoughtbot/bamboo/commit/1bfad10586633fe14c7f7f71332c3056a8a002e3 and https://github.com/thoughtbot/bamboo/commit/5ad6870c8523671bc67d52d32c46cc5ccfb9aa47

paulcsmith:
@optikfluffel That's a really good idea!

PR #224, opened 2016-10-25
Adds general Bamboo.ApiError and updates adapters to use it

### LINKED ISSUE #196: Need a common Exception type
Adding this for discussion mostly.

Each of the individual Adapters defines its own Exception module/struct. That means if you want to catch the exceptions thrown by `deliver_now` you have to use one specific to the Adapter that's currently configured for use. That makes `Bamboo.Mailer` a leaky abstraction since you have to change the error handling if you change the adapter.

e.g.:

```
try do
  Email.welcome_email |> Mailer.deliver_now
rescue
  e in Bamboo.MailgunAdapter.ApiError ->
    # ... record the error
end
```

Having a common `Bamboo.Mailer.Error` that the adapters could use would solve that problem. But, maybe it would be better to have a more Elixir style "maybe monad" return using `{:ok, response}` and `{:error, response}` instead of raising an exception?

jbernardo95:
This PR resolves #196.
- Removes specific api errors from adapters with external api calls
- Creates a common `Bamboo.ApiError` module
- Adds a `raise_api_error` function to every adapter with external api calls

### DIFF
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -28,31 +28,7 @@ defmodule Bamboo.SendGridAdapter do
   @behaviour Bamboo.Adapter
 
   alias Bamboo.Email
-
-  defmodule ApiError do
-    defexception [:message]
-
-    def exception(%{message: message}) do
-      %ApiError{message: message}
-    end
-
-    def exception(%{params: params, response: response}) do
-      filtered_params = params |> Plug.Conn.Query.decode |> Map.put("key", "[FILTERED]")
-
-      message = """
-      There was a problem sending the email through the SendGrid API.
-
-      Here is the response:
-
-      #{inspect response, limit: :infinity}
-
-      Here are the params we sent:
-
-      #{inspect filtered_params, limit: :infinity}
-      """
-      %ApiError{message: message}
-    end
-  end
+  alias Bamboo.ApiError
 
   def deliver(email, config) do
     api_key = get_key(config)
@@ -61,11 +37,11 @@ defmodule Bamboo.SendGridAdapter do
 
     case :hackney.post(url, headers(api_key), body, [:with_body]) do
       {:ok, status, _headers, response} when status > 299 ->
-        raise(ApiError, %{params: body, response: response})
+        raise_api_error(body, response)
       {:ok, status, headers, response} ->
         %{status_code: status, headers: headers, body: response}
       {:error, reason} ->
-        raise(ApiError, %{message: inspect(reason)})
+        raise_api_error(inspect(reason))
     end
   end
 
@@ -83,6 +59,34 @@ defmodule Bamboo.SendGridAdapter do
       nil -> raise_api_key_error(config)
       key -> key
     end
+  end
+
+  defp raise_api_error(message), do: raise(ApiError, message: message)
+  defp raise_api_error(params, response) do
+    filtered_params = params |> Plug.Conn.Query.decode |> Map.put("key", "[FILTERED]")
+
+    message = """
+    There was a problem sending the email through the SendGrid API.
+
+    Here is the response:
+
+    #{inspect response, limit: :infinity}
+
+    Here are the params we sent:
+
+    #{inspect filtered_params, limit: :infinity}
+
+    If you are deploying to Heroku and using ENV variables to handle your API key,
+    you will need to explicitly export the variables so they are available at compile time.
+    Add the following configuration to your elixir_buildpack.config:
+
+    config_vars_to_export=(
+      DATABASE_URL
+      SENDGRID_API_KEY
+    )
+    """
+
+    raise(ApiError, message: message)
   end
 
   defp raise_api_key_error(config) do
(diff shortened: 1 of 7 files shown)


lib/bamboo/adapters/sendgrid_adapter.ex
```diff
@@ -89,6 +56,34 @@ defmodule Bamboo.SendgridAdapter do
     end
   end
 
+  defp raise_api_error(message), do: raise(ApiError, message: message)
+  defp raise_api_error(params, response) do
+    filtered_params = params |> Plug.Conn.Query.decode |> Map.put("key", "[FILTERED]")
+
+    message = """
+    There was a problem sending the email through the SendGrid API.
```
drapergeek: What do you think about extracting all of these info the `ApiError` itself? Passing in the params (pre-filtered or parsed), response, name of the adapter and some additional information would remove some of this duplication.

I think it also provides a nice template that other adapters can follow without any additional work by just passing in the params and response.

drapergeek:
Thanks @jbernardo95 sorry for the delay on this. I left a comment, I'd love to see this moved forward and I think it's pretty close.

paulcsmith:
Yes I definitely want to get this in, but was going to merge for 1.0, which is (hopefully) going to be released soon!

lib/bamboo/adapters/sendgrid_adapter.ex
jbernardo95: @drapergeek Makes a lot more sense than the current implementation, I will improve this and let you know when it is done.

jbernardo95:
@paulcsmith @drapergeek Moved the `raise_api_error` function to the `ApiError` module. What do you think ?

lib/bamboo/api_error.ex
```diff
@@ -0,0 +1,34 @@
...
+  @moduledoc """
+  Error used to represent a problem when sending emails through an external email service API.
+  """
+
+  defexception [:message]
+
+  def raise_api_error(message), do: raise(__MODULE__, message: message)
+  def raise_api_error(adapter, response, params, extra_message \\ "") when is_atom(adapter) do
+    message = """
+    There was a problem sending the email through the #{adapter_name(adapter)} API.
```
drapergeek: I may be the only one who thinks this but it seems weird to pass in the module itself and call a function back on it here. What do you think about just passing in the name directly? With this change, the `ApiError` now has knowledge of a function on the module passed in which seems like information it shouldn't really care about.

jbernardo95: @derekprior I though about this too, but for me it also seems wired to have a string, which is in some way related to what the module is, as the first argument of the function (`raise_api_error("Mailgun", ...)`)
What do you think of having a `@service_name` attribute on every adapter module and passing it to the function like so `raise_api_error(@service_name, ...)` ?

drapergeek: I think that's a great solution.

You may also consider a map or keyword list so that the arguments can be named which would make it more clear.

jbernardo95:
@drapergeek Updated ^

paulcsmith:
@jbernardo95 This looks great! I just merged https://github.com/thoughtbot/bamboo/commit/00b24cc315009c041377d6bc9d126ebd6d1caa7e which caused some issues because of the now obsolete Heroku error messages. Could you remove the Heroku messages and rebase? Once you do, please ping me and I'll merge this in. Thanks so much!

But if you're busy, let me know and I can fix it up sometime

jbernardo95:
@paulcsmith Fixed 😉

lib/bamboo/adapters/send_grid_adapter.ex
```diff
@@ -23,36 +23,23 @@ defmodule Bamboo.SendGridAdapter do
...
+  @extra_api_error_message """
+  If you are deploying to Heroku and using ENV variables to handle your API key,
+  you will need to explicitly export the variables so they are available at compile time.
+  Add the following configuration to your elixir_buildpack.config:
+
+  config_vars_to_export=(
+    DATABASE_URL
+    SENDGRID_API_KEY
+  )
+  """
```
paulcsmith: Could you remove these messages as they are no longer necessary (see PR #275 for more info)

paulcsmith:
That was fast! Just one tiny thing and this is good to go!

jbernardo95:
@paulcsmith Too fast to be perfect x)

Should be fixed now. Let me know If I should remove the `extra_message` parameter in the `raise_api_error` function.

lib/bamboo/api_error.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Bamboo.ApiError do
+  @moduledoc """
+  Error used to represent a problem when sending emails through an external email service API.
+  """
+
+  defexception [:message]
+
+  def raise_api_error(message), do: raise(__MODULE__, message: message)
```
paulcsmith: Can you add docs for this, alone with an example for setting the `extra_message`? I think that could be helpful if an adapter wants to add a bit more details

paulcsmith:
Good point about the extra_message. I think we should add docs and leave it in. Nice work on this!

jbernardo95:
@paulcsmith Documentation check !

I also added a small check so that when no `extra_message` is provided the error message does not appear with two blank lines at the end.

paulcsmith:
Very nice docs and nice catch with removing the extra newlines 👍

PR #225, opened 2016-10-26
Print headers in email preview

Linuus:
This PR prints the headers of the email to the preview.

I added this because I wanted to see things like "Reply-To" headers in the preview.

Is this something you'd be interested in?

Example:

screen

### DIFF
--- a/lib/bamboo/plug/helper.ex
+++ b/lib/bamboo/plug/helper.ex
@@ -17,6 +17,12 @@ defmodule Bamboo.EmailPreviewPlug.Helper do
     |> Enum.join(", ")
   end
 
+  def format_headers(values) when is_binary(values), do: values
+  def format_headers(values) when is_list(values) do
+    Enum.join(values, ", ")
+  end
+  def format_headers(values), do: inspect(values)
+
   def format_text(nil), do: ""
   def format_text(text_body) do
     String.replace(text_body, "\n", "<br>")
--- a/lib/bamboo/plug/index.html.eex
+++ b/lib/bamboo/plug/index.html.eex
@@ -122,11 +122,13 @@
         font-weight: 700;
       }
 
-      .email-preview-recipients {
+      .email-preview-recipients,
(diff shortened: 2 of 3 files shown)


lib/bamboo/plug/helper.ex
```diff
@@ -17,6 +17,12 @@ defmodule Bamboo.EmailPreviewPlug.Helper do
     |> Enum.join(", ")
   end
 
+  def format_values(values) when is_binary(values), do: values
+  def format_values(values) when is_list(values) do
+    Enum.join(values, ", ")
+  end
+  def format_values(values), do: inspect(values)
```
paulcsmith: How about calling this `format_headers`?

paulcsmith: Also, what is this catch all function for? I think headers will always be a list or a string, or is there something else that it could be?

paulcsmith:
This is a great idea!

lib/bamboo/plug/helper.ex
Linuus: I actually called it `format_headers` first, but then I realized it's not actually doing anything header specific. It just takes a string or list and formats it. So, it can be used for other stuff as well.

But sure, I can rename it if you want, just let me know and I'll fix it :)

The catch all was just because I don't actually know if there can be other values. Can it be a map or tuple? If it can't it's not needed of course :)

paulcsmith: I think for now `format_headers` is good and it can be renamed `format_values` later if it's needed in other parts of the library. We can leave the catch all just in case, but it would be good to add a test case of it

Linuus: Sure! I'll fix it!

Linuus:
@paulcsmith Thank you for the quick review! I've updated the PR.

paulcsmith:
@Linuus You're welcome. Thanks for the PR!

PR #235, opened 2016-12-12
Adds support to send email headers in MailgunAdapter

### LINKED ISSUE #232: Support for mail headers in Mailgun adapter
According to [Mailgun documentation](https://documentation.mailgun.com/api-sending.html#sending) one can set custom mail headers by adding a field prefixed with `h:` in the HTTP request:

> h:X-My-Header 	h: prefix followed by an arbitrary value allows to append a custom MIME header to the message (X-My-Header in this case). For example, h:Reply-To to specify Reply-To address.

With Bamboo we set headers with the `Email.put_header/3` function. They are stored in the `headers` map of the `Email` struct. But the Mailgun adapter doesn't make use of them at the moment.

Would it be possible to implement that? It doesn't seem complicated.

jbernardo95:
Fixes #232.

### DIFF
--- a/test/lib/bamboo/adapters/mailgun_adapter_test.exs
+++ b/test/lib/bamboo/adapters/mailgun_adapter_test.exs
@@ -80,13 +80,14 @@ defmodule Bamboo.MailgunAdapterTest do
     assert request_path == "/test.tt/messages"
   end
 
-  test "deliver/2 sends from, html and text body, subject, and headers" do
+  test "deliver/2 sends from, subject, text body, html body and headers" do
     email = new_email(
       from: "from@foo.com",
       subject: "My Subject",
       text_body: "TEXT BODY",
       html_body: "HTML BODY",
     )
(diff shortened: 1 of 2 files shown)


test/lib/bamboo/adapters/mailgun_adapter_test.exs
```diff
@@ -96,8 +97,9 @@ defmodule Bamboo.MailgunAdapterTest do
     assert params["subject"] == email.subject
     assert params["text"] == email.text_body
     assert params["html"] == email.html_body
+    assert params["h:X-My-Header"] == "my_header_value" 
 
-    hashed_token  = Base.encode64("api:" <> @config.api_key)
+    hashed_token = Base.encode64("api:" <> @config.api_key)
```
paulcsmith: Nice catch

paulcsmith:
Thank you!

jbernardo95:
Welcome !

PR #239, opened 2016-12-27
Fixes emails module name inconsistency in README.md

### LINKED ISSUE #238: MyApp.Email vs MyApp.Emails in README
I noticed that README is using both `MyApp.Email` and `MyApp.Emails` - do you have a recommendation how to name the module? Obviously it's up to the user but I think it would be good to be consistent.

PS one merit of calling it `MyApp.Emails` is that it won't clash with `MyApp.Email` schema if one would ever need it (e.g. user could have multiple email addresses)

jbernardo95:
Fixes #238.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ defmodule MyApp.Mailer do
 end
 
 # Define your emails
-defmodule MyApp.Email do
+defmodule MyApp.Emails do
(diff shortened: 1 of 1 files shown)


paulcsmith:
Thank you!

PR #241, opened 2017-01-05
Fix HashDict deprecations for Map (elixir 1.4)

ddrmanxbxfr:
This PR removes the deprecations for HashDict Usage in favor of Map.

HashDict has been deprecated since Elixir 1.2 as Erlang 1.8 got a huge performance boost on maps.

Map is now the favored data structure for maps.

See :

https://github.com/elixir-lang/elixir/blob/v1.2.0/CHANGELOG.md#erlang-18-support
https://hexdocs.pm/elixir/Map.html

Thanks!


paulcsmith:
Thank you very much!

PR #245, opened 2017-01-11
Change Email to Emails so example function call match module name

andreaseriksson:


paulcsmith:
Good catch! Thanks @andreaseriksson

PR #246, opened 2017-01-15
Adds deliver_now/2 and deliver_later/2 with possibility to pass sender credentials dynamically

mAlishera:


zolrath:
I was just exploring the docs looking for this exact feature! Looking forward to this, thanks for implementing it!

mAlishera:
Hope it will be accepted, thanks for your feedback, @zolrath 😃

mAlishera:
ping @paulcsmith

paulcsmith:
@mAlishera Thank you for the ping! Sometimes I get a bit behind so it's helpful.

I think what I would prefer to see for this feature is adapter's implementing this with adapter specific helpers. So you would do something like this

```elixir
new_email(to: "someone", from: "someone", etc.)
# This will put the sender credentials in the `private` field of the `Bamboo.Email` struct
|> WhateverEmailService.Helpers.put_sender_credentials(my_credentials)
|> Mailer.deliver_now
```

Then the `WhateverEmailServiceAdapter` would use the sender credentials from the email when delivering. You can see how to implement adapter specific functionality here https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/send_grid_helper.ex#L31 and  https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/send_grid_adapter.ex#L170.

This way the mailer API is kept clean and simple (doesn't mix in email provider specific concerns). It's also easier to test because the email itself knows which credentials to use. 

What do you think of that idea? Would you be willing to give that a shot? If you need help or need someone else to tackle this, let me know! I'll go ahead and close this in the meantime.

mAlishera:
@paulcsmith Thank you for your response! 
I'll try to implement it the way you've described. 
So actually using it with Bamboo.SMTPAdapter I need to add my changes via MR to fewlinesco/bamboo_smtp, right?

paulcsmith:
@mAlishera Yes that's correct. You'd have to add a `Bamboo.SMTPHelper` on @fewlinesco's repo and change the adapter to optionally use the passed in sender credentials. You may want to open an issue on their repo first to make sure it's something they would accept though.

Good luck! Ping me if you need anything

Issue #249, opened 2017-01-26
LocalAdapter doesn't work on Heroku


jsteiner:
I get the following error when using the `LocalAdapter` on Heroku:

```
Request: GET /sent_emails
2017-01-18T01:15:26.895034+00:00 app[web.1]: ** (exit) an exception was raised:
2017-01-18T01:15:26.895035+00:00 app[web.1]:     ** (File.Error) could not read file "/tmp/build_22154930305f53f60cc455254f4e2498/deps/bamboo/lib/bamboo/plug/no_emails.html.eex": no such file or directory
```

My best guess at what's happening here is that the path get's expanded for `__DIR__` [here](https://github.com/thoughtbot/bamboo/blob/a41ce9cc2c5639e830268a33268625fca669ef59/lib/bamboo/plug/email_preview_plug.ex#L74) at compile time, then Heroku moves the built app to another directory.

I tried to fix this a few different ways with relative paths, but they were trying to load relative to the application directory, not the (bamboo) package directory. I was able to get it to work by loading it from the `deps/bamboo/` dir in the app directory, but that doesn't seem like a reliable solution.

I haven't dug into it yet, but I wonder if we could use [`compile_file/2`](https://hexdocs.pm/eex/EEx.html#compile_file/2) to compile the template to a function and pass it arguments at runtime. It looks like [`function_from_file/5`](https://hexdocs.pm/eex/EEx.html#function_from_file/5) wants the template arguments at compile time, so we can't use that.

jsteiner:
Actually, I think I misunderstood how `function_from_file/5` works. It looks like the array of args just takes the names of the arguments. Not the values.

paulcsmith:
Closed by #250

PR #250, opened 2017-01-27
Get LocalAdapter working on Heroku

jsteiner:
Previously, the LocalAdapter wasn't working on Heroku, failing at
runtime with:

```
Request: GET /sent_emails
2017-01-18T01:15:26.895034+00:00 app[web.1]: ** (exit) an exception was
raised:
2017-01-18T01:15:26.895035+00:00 app[web.1]:     ** (File.Error) could
not read file
"/tmp/build_22154930305f53f60cc455254f4e2498/deps/bamboo/lib/bamboo/plug/no_emails.html.eex":
no such file or directory
```

It seems that the path was being expanded for `__DIR__` at compile time,
then subsequently moved the app directory, causing the lookup to fail.

This solves the issue by converting the template to a function at
compile time, so we don't have to do any runtime file lookups.

### DIFF
--- a/lib/bamboo/plug/email_preview_plug.ex
+++ b/lib/bamboo/plug/email_preview_plug.ex
@@ -1,6 +1,16 @@
 defmodule Bamboo.EmailPreviewPlug do
   use Plug.Router
+  require EEx
   alias Bamboo.SentEmail
+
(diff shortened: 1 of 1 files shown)


jsteiner:
Addresses https://github.com/thoughtbot/bamboo/issues/249

paulcsmith:
👍

PR #260, opened 2017-03-02
Initial attempt at supporting different matchers

rawkode:
Signed-off-by: David McKay <david@rawkode.com>

I don't believe this is optimal, I'm really looking for some guidance.

I believe `do_match` should be `defp`, but I don't know if that is possible because `assert_delivered_with` is a macro

### DIFF
--- a/test/lib/bamboo/adapters/test_adapter_test.exs
+++ b/test/lib/bamboo/adapters/test_adapter_test.exs
@@ -148,6 +148,20 @@ defmodule Bamboo.TestAdapterTest do
     assert_delivered_with text_body: ~r/like/
   end
 
+  test "ensure assert_delivered_with regex matching doesn't provide a false positive" do
+    new_email(to: {nil, "foo@bar.com"}, from: {nil, "foo@bar.com"}, text_body: "I really like coffee")
+      |> TestMailer.deliver_now
+
+    try do
+      assert_delivered_with text_body: ~r/tea/
+    rescue
+      error in [ExUnit.AssertionError] ->
+          assert error.message =~ "do not match"
+    else
+        _ -> flunk "assert_delivered_with should have failed"
+    end
+  end
+
   test "assert_no_emails_delivered shows the delivered email" do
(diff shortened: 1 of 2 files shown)


test/lib/bamboo/adapters/test_adapter_test.exs
```diff
@@ -141,6 +141,13 @@ defmodule Bamboo.TestAdapterTest do
     end
   end
 
+  test "assert_delivered_with allows regex matching" do
+    new_email(to: {nil, "foo@bar.com"}, from: {nil, "foo@bar.com"}, text_body: "I really like coffee")
+      |> TestMailer.deliver_now
+
+    assert_delivered_with text_body: ~r/like/
```
paulcsmith: Could you also add something similar to https://github.com/thoughtbot/bamboo/pull/260/files#diff-162d78c24e2d8c7dfed124276ea1f3d6L114 that tests for emails that'd don't have a matching body? That way the tests are a bit more reliable

rawkode: Added :+1:

test/lib/bamboo/adapters/test_adapter_test.exs
```diff
@@ -141,6 +141,27 @@ defmodule Bamboo.TestAdapterTest do
...
+    new_email(to: {nil, "foo@bar.com"}, from: {nil, "foo@bar.com"}, text_body: "I really like coffee")
+      |> TestMailer.deliver_now
+
+    try do
+      assert_delivered_with text_body: ~r/tea/
+    rescue
+      error in [ExUnit.AssertionError] ->
+          assert error.message =~ "do not match"
+    else
+        _ -> flunk "assert_delivered_with should have failed"
```
paulcsmith: This line and the one above are indented 2 spaces too much. Could you unindent them?

paulcsmith:
I just realized it would probably be helpful to update the docs so people know about this. I'll add an issue so this can be done in a separate PR. Thanks for the awesome work @rawkode

PR #267, opened 2017-03-21
fix example

chenxsan:
I think the original example is wrong.


paulcsmith:
Both versions would work, but this makes more sense. Good catch and thanks @chenxsan

PR #269, opened 2017-03-24
Update EmailPreviewPlug links

jergason:
The README.md currently links to a nonexistent module for email previewing. This fixes the link.


paulcsmith:
@jergason Thanks for the PR! This was due to a bit of confusion about the README. The README on master is for an unpublished version of Bamboo. Once we are ready for 1.0 and push to hex, the README will work correctly. Instead I've added a note to the README (#279) to direct people to hex to get the docs for the latest *published* version of Bamboo. The README here is for master only. 

I hope that makes sense!

PR #273, opened 2017-04-04
Create New Sparkpost Adapter

genericlady:
Create a new SparkPost adapter that supports the most recent version of bamboo, elixir 1.4 and phoenix 1.3.

Emails send fine using SparkPost Elixir library but I really want to continue using Bamboo. bamboo_sparkpost has some downsides to it's configuration and I feel better making my own adapter from a forked copy of this project and debugging it along the way.

This way I can write my own tests and make sure it all works as it should.

### DIFF
--- a/lib/bamboo/adapters/spark_post_adapter.ex
+++ b/lib/bamboo/adapters/spark_post_adapter.ex
@@ -1,0 +1,147 @@
+defmodule Bamboo.SparkPostAdapter do
+  @doc """
+  Sends email using the Spark Post API.
+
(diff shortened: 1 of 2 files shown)


paulcsmith:
Hi @genericlady, thank you so much for writing this. I'm trying to keep the core of Bamboo smaller, so you could extract a library for this, publish to hex, and then I can link to it in the README? I know there already is one so if the other one doesn't work, then I can link to yours instead

Let me know if this doesn't make sense or you want any help with it!

PR #275, opened 2017-04-09
Remove obsolete heroku instructions

tsubery:
Since this [commit](https://github.com/HashNuke/heroku-buildpack-elixir/commit/6251590e3a354d4b8ac99960977daf417d3128db) the buildpack exports all environment variables by default so this section is unnecessary


paulcsmith:
Nice catch! Thanks @tsubery

PR #277, opened 2017-04-12
Display Sender's name on the preview page.

linjunpop:
## Before
image

## After
image


dbernheisel:
Thanks @linjunpop this looks like it works for me locally. I was actually just on a project where I formatted the `To` field, but was a little surprised when I didn't see it in the sent emails viewer.

What do you think of adding a space in between the name and the address?
ie, from `My Name<email@example.com>`  to `My Name <email@example.com>`. I think this is how email programs usually display this.

PR #283, opened 2017-04-24
docs(README.md): Fix the url for the EmailPreviewPlug

tomciopp:
This PR address a broken link to the EmailPreviewPlug


paulcsmith:
@tomciopp Thanks for the PR! The README is following master so some links are broken until the new version is shipped. 1.0rc1 is almost done, and once it is the link will work again. So we'll leave this for now since the best way to get docs for the latest published version is to go to hex.

I know it's not ideal, but it's the best we can do right now

PR #284, opened 2017-05-05
Raise if adapter does not support attachments yet

paulcsmith:


paulcsmith:
Not all adapters will support attachments right away. This makes it so that if you try to add an attachment, but are using an adapter that doesn't do anything with them, it raises letting you know that's the problem. Otherwise it would be successful, but the attachments would do nothing

Issue #286, opened 2017-05-08
Change attachment struct to support file data instead of filename


sbrink:
I propose that we change the current `Attachement` struct before releasing version 1.0.

**Why?**

At the moment an attachment struct looks like this:

```
defstruct filename: nil, content_type: nil, path: nil
```

This only allows uploads from the file system, not dynamically generated contents.
An use case for generated contents would be `.ics` files. You want to append a 
generated `.ics` file for adding an event to the user's calendar.

Furthermore every adapter has to implement the correct encoding - this is redundant.
Here's an example I found from a patch for the bamboo_stmp adapter:

```
defp add_attachment_body(body, attachment) do
  file = File.read!(attachment)
   "#{body}\r\n#{Base.encode64(file)}\r\n"
 end
```

**Proposal:**

We change the struct definition to:

```
defstruct filename: nil, content_type: nil, data: nil
```

We could:

1. Split the `new` function into  `new_from_data` and `new_from_attachment`
2. Use pattern matching in the `new` function

The new new method does the job of file reading and encoding.

What do you think?

schurig:
+1

In most of my use cases I would use dynamically generated data that I wouldn't want to store on the hard disk before attaching them to outgoing emails. I think it makes sense to change the `Attachment` struct before releasing version 1.0 as @sbrink proposes.

paulcsmith:
Thank you so much for the suggestion.

I like this idea. I like option 1 of using two different functions, but I think `new_from_file` might make more sense since they are both attachments. 

I'm not sure about doing the Base64 encoding outside the adapter because there may be adapters that don't need that. 

Would one of you be willing to work on a PR for this?

andrewtimberlake:
I agree with this.
I made a change in my adapter to accept `{:binary, data}` in the path to allow for this (https://github.com/andrewtimberlake/bamboo_sparkpost/blob/feature/attachments/lib/bamboo/adapters/sparkpost_adapter.ex#L173)

sbrink:
I had no time last week. @schurig and I are working on it now.

paulcsmith:
You rock. Thank you!

schurig:
Hi everyone!
@sbrink and I were working on a version that

* [supports `|> put_attachment %Attachment{filename: "event.ics", data: "raw content.."}`](https://github.com/workshops-de/bamboo/blob/76779e9b7e97ec57278a2ffba8d45b805d09af7b/lib/bamboo/email.ex#L223-L225)
[...]

PR #292, opened 2017-05-23
Attachments: Support file data in attachment struct + mailgun attachment support

schurig:
As discussed in #286 @sbrink and I were working on an improved version of the attachment support:

### Changes

* supports `|> put_attachment %Attachment{filename: "event.ics", data: "raw content..", content_type: "text/calendar"}`
* sets `Attachment.data` when using `|> put_attachment "/hi.jpg"` or `|> put_attachment %Plug{..}`
* supports attachments in the mailgun and test adapter
* updates the mandrill adapter to use the `attachment.data`

### Todo

* tests for the mailgun adapter (attachment support)

### DIFF
--- a/lib/bamboo/adapters/mandrill_adapter.ex
+++ b/lib/bamboo/adapters/mandrill_adapter.ex
@@ -108,11 +108,11 @@ defmodule Bamboo.MandrillAdapter do
   defp attachments(%{attachments: attachments}) do
     attachments
     |> Enum.reverse
-    |> Enum.map(fn(att) ->
+    |> Enum.map(fn(attachment) ->
       %{
-        name: att.filename,
-        type: att.content_type,
-        content: Base.encode64(File.read!(att.path))
+        name: attachment.filename,
+        type: attachment.content_type,
+        content: Base.encode64(attachment.data)
       }
     end)
   end
--- a/lib/bamboo/adapters/test_adapter.ex
+++ b/lib/bamboo/adapters/test_adapter.ex
@@ -52,4 +52,7 @@ defmodule Bamboo.TestAdapter do
   def clean_assigns(email) do
     %{email | assigns: :assigns_removed_for_testing}
   end
+
+  @doc false
+  def supports_attachments?, do: true
 end
--- a/lib/bamboo/attachment.ex
+++ b/lib/bamboo/attachment.ex
@@ -2,16 +2,16 @@ defmodule Bamboo.Attachment do
   @moduledoc """
   """
 
-  defstruct filename: nil, content_type: nil, path: nil
+  defstruct filename: nil, content_type: nil, path: nil, data: nil
 
   @doc ~S"""
   Creates a new Attachment
 
   Examples:
-    Attachment.new("/path/to/attachment.png")
-    Attachment.new("/path/to/attachment.png", filename: "image.png")
-    Attachment.new("/path/to/attachment.png", filename: "image.png", content_type: "image/png")
-    Attachment.new(params["file"]) # Where params["file"] is a %Plug.Upload
+    Bamboo.Attachment.new("/path/to/attachment.png")
+    Bamboo.Attachment.new("/path/to/attachment.png", filename: "image.png")
+    Bamboo.Attachment.new("/path/to/attachment.png", filename: "image.png", content_type: "image/png")
+    Bamboo.Attachment.new(params["file"]) # Where params["file"] is a %Plug.Upload
   """
   def new(path, opts \\ [])
   if Code.ensure_loaded?(Plug) do
@@ -21,7 +21,8 @@ defmodule Bamboo.Attachment do
   def new(path, opts) do
     filename = opts[:filename] || Path.basename(path)
     content_type = opts[:content_type] || determine_content_type(path)
-    %__MODULE__{path: path, filename: filename, content_type: content_type}
+    data = File.read!(path)
+    %__MODULE__{path: path, data: data, filename: filename, content_type: content_type}
   end
 
   defp determine_content_type(path) do
--- a/lib/bamboo/email.ex
+++ b/lib/bamboo/email.ex
@@ -218,7 +218,13 @@ defmodule Bamboo.Email do
       #...
     end
   """
-  def put_attachment(%__MODULE__{attachments: attachments} = email, %Attachment{filename: _filename, data: _data} = attachment) do
+  def put_attachment(%__MODULE__{attachments: attachments} = email, %Attachment{filename: nil}) do
+    raise "You must provide a filename for the attachment."
+  end
+  def put_attachment(%__MODULE__{attachments: attachments} = email, %Attachment{data: nil}) do
+    raise "The attachment must contain data."
+  end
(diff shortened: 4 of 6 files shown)


randycoulman:
How close is this PR to being ready to merge?  I'm working on a project that needs to generate and attach vCards to e-mails, so being able to create an attachment from data rather than a file would be super helpful.

We use the SendGrid adapter, so I'm starting to work on a PR to add attachment support for that adapter today.  I'd love it if I could base my work off of the code in this PR.

Thanks for all your work on this library!

lib/bamboo/adapters/mandrill_adapter.ex
```diff
@@ -112,7 +112,7 @@ defmodule Bamboo.MandrillAdapter do
       %{
         name: att.filename,
         type: att.content_type,
-        content: Base.encode64(File.read!(att.path))
+        content: Base.encode64(att.data)
```
paulcsmith: While we're in this PR, could you rename `att` to `attachment`?

lib/bamboo/adapters/test_adapter.ex
```diff
@@ -52,4 +52,7 @@ defmodule Bamboo.TestAdapter do
   def clean_assigns(email) do
     %{email | assigns: :assigns_removed_for_testing}
   end
+
+  @doc false
+  def supports_attachments?, do: true
```
paulcsmith: Good catch!

lib/bamboo/attachment.ex
```diff
@@ -2,26 +2,27 @@ defmodule Bamboo.Attachment do
...
 
   @doc ~S"""
   Creates a new Attachment
 
   Examples:
-    Attachment.new("/path/to/attachment.png")
-    Attachment.new("/path/to/attachment.png", filename: "image.png")
-    Attachment.new("/path/to/attachment.png", filename: "image.png", content_type: "image/png")
-    Attachment.new(params["file"]) # Where params["file"] is a %Plug.Upload
+    Attachment.new_from_file("/path/to/attachment.png")
```
paulcsmith: Can you change these to `Bamboo.Attachment` so it's clear this is coming from Bamboo?

lib/bamboo/email.ex
```diff
@@ -220,7 +220,10 @@ defmodule Bamboo.Email do
       #...
     end
   """
```
paulcsmith: I think adding docs for sending data would be awesome here

lib/bamboo/attachment.ex
paulcsmith: Also I think `new` is better here since there isn't a second function. Alternatively you could add `new_from_data` so there were two different functions. I'd probably prefer just `new` though

randycoulman:
I tried this branch on something I was working on, and I noticed there were a number of test failures that will need to be resolved.  It doesn't look like CircleCI is running builds right now, so there's no indication of a problem here.

From the test failures, I couldn't tell if the code was doing the right thing and the tests needed to be changed, or if the tests were right and the code was wrong.  I suspect the latter, but I'm not confident enough in that to be able to suggest fixes.

paulcsmith:
Yeah I'm not sure what's up with Circle. I'll work on that. Busy with client work now, but hopefully in the next few days. In the meantime I can run it locally before merging to make sure it all works ok. @schurig are you seeing failures locally?

schurig:
I'll do it asap! I remember that I fixed all tests. But I can't confirm at the moment.

schurig:
I haven't found a way to test the attachments for the mailgun adapter. Can anyone help me with it? 

It uses [multipart as content type](https://github.com/workshops-de/bamboo/blob/develop/lib/bamboo/adapters/mailgun_adapter.ex#L158-L164) and I can't get the attachment by calling `assert_receive {:fake_mailgun, struct}` in the test.

paulcsmith:
@schurig I just came to check in and didn't realize you left that comment. Sorry about that. I would remove the changes from the MailgunAdapter and it's tests and instead do a separate PR for those. That way we can merge in the attachments changes and work on the adapters separately. Does that sound good?

schurig:
@paulcsmith that sounds good!!

schurig:
@paulcsmith ready for a 2nd review.

lib/bamboo/email.ex
```diff
@@ -203,7 +203,27 @@ defmodule Bamboo.Email do
...
+
+  ## Example
+    def create(conn, params) do
+      #...
+      email
+      |> put_attachment(%Bamboo.Attachment{filname: "event.ics", data: "BEGIN:VCALENDAR..."})
+      #...
+    end
+  """
+  def put_attachment(%__MODULE__{attachments: attachments} = email, %Attachment{filename: _filename, data: _data} = attachment) do
```
paulcsmith: I believe you can shorten the pattern match of the attach meant to `%Attachment{}` since you wouldn't be able to initialize that struct without those fields

schurig: You actually can. Then the values are set to `nil` because we don't specify anything as enforced keys:

```
iex(1)> %Bamboo.Attachment{}
%Bamboo.Attachment{content_type: nil, data: nil, filename: nil, path: nil}
```

But you're right - since we don't check the actual value it doesn't make sense for that check. But it would only make sense to put pass an Attachment struct with both keys (`filename` and `data`) specified in there. Do you think we should throw an error if one of it is empty?

paulcsmith: Maybe we have 2 functions. One for the happy path and another for attachments that are missing data and filename

```elixir
def put_attachment(_email, %Attachment{filename: nil, data: nil} = attachment) do
  raise "You must pass a valid attachment, instead got: #{inspect attachment}"
end

def put_attachment(%__MODULE__{attachments: attachments} = email, %Attachment{} = attachment) do
  # happy path!
end
```

I think this would cover the common error case. We can add different error cases later if people run into things often. Thoughts?

schurig: @paulcsmith I've made the error messages more specific since it would still jump into the happy path if one of the params is `nil`. What do you think?

lib/bamboo/email.ex
```diff
@@ -218,7 +218,13 @@ defmodule Bamboo.Email do
       #...
     end
   """
-  def put_attachment(%__MODULE__{attachments: attachments} = email, %Attachment{filename: _filename, data: _data} = attachment) do
+  def put_attachment(%__MODULE__{attachments: attachments} = email, %Attachment{filename: nil}) do
+    raise "You must provide a filename for the attachment."
+  end
+  def put_attachment(%__MODULE__{attachments: attachments} = email, %Attachment{data: nil}) do
+    raise "The attachment must contain data."
```
paulcsmith: Can you include the attachment in both error message so that it's easier to see what they passed in? 

```elixir
raise "The attachment must contain data, instead got: #{inspect attachment}"
```

schurig: Good point! Done. Also cleaned up a bit. There were some warnings about not used variables.

paulcsmith:
Wonderful work! Thanks for your first contribution to Bamboo!

Issue #280, opened 2017-04-17
Getting "Unexpected error" with Sendgrid Adapter


luccasmaso:
I have created and tested my sendgrid api_key via curl and it works with directly with api v3. With the lib I've tried some other examples but I don't know what else I can do/test because the error is too generic. 

Thanks

```
elixir_version=1.4.2
```

My `prod.exs` config:
```
config :my_app, MyApp.Mailer,
  adapter: Bamboo.SendgridAdapter,
  api_key: System.get_env("SENDGRID_API_KEY")
```

Code:
```
defmodule MyApp.Mailer do
  use Bamboo.Mailer, otp_app: :my_app
end

...

import Bamboo.Email

mail =
    new_email() 
    |> from("no-reply@my.app") 
    |> to("some@email.com") 
    |> subject("Title") 
    |> text_body("text")

mail |> MyApp.Mailer.deliver_now
```

```
%Bamboo.Email{assigns: %{}, bcc: [], cc: [], from: {nil, "no-reply@my.app"}, headers: %{}, html_body: nil, private: %{}, subject: "Title", text_body: "text", to: [nil: "some@email.com"]}

** (Bamboo.SendgridAdapter.ApiError) There was a problem sending the email through the SendGrid API.

Here is the response:

"{\"errors\":[\"Unexpected error\"],\"message\":\"error\"}"

Here are the params we sent:

%{"from" => "no-reply@my.app", "fromname" => "", "key" => "[FILTERED]", "subject" => "Title", "text" => "Sample text", "to" => ["some@email.com"]}
```

luccasmaso:
?

paulcsmith:
I would contact SendGrid support or post on the Elixir forums. That error SendGrid is returning seems very cryptic so I'm not sure what the problem is. It could be that it doesn't like the fake email, but I'm not sure

sebbean:
having same error

coryt:
I also get this error. I've narrowed it down to something in the substitutions but I'm not sure what is the cause.

> Here is the response:
> 
> "{\"errors\":[{\"message\":\"Bad Request\",\"field\":null,\"help\":null}]}"
> 
> Here are the params we sent:
> 
[...]

PR #293, opened 2017-05-31
Convert send_grid_adapter to SendGrid API v3

randycoulman:
The existing `SendGridAdapter` uses v2 of the SendGrid API.  In https://sendgrid.com/docs/API_Reference/Web_API/mail.html, it recommends migrating to the new v3 API:

> To current v2 users: We encourage migrating to the new endpoint to take advantage of upcoming features.

I was looking at adding attachment support to the existing adapter, and found that it would be much easier to do with the v3 API instead.  So, I'm submitting this PR first, and will then add attachment support on top of it.

For reference, the v3 API docs are at https://sendgrid.com/docs/API_Reference/api_v3.html.

I've done my best to preserve the existing API for both the adapter and the helper, but I've changed the internal representation of the private data that the helper adds to the e-mail.  The new format is simpler and less coupled to the resulting API format:

```
%{
  send_grid_template: %{
    template_id: "YOUR ID HERE",
    substitutions: %{
      "%foo%" => "bar"
    }
  }
}
```

I've tested this adapter in an existing application and it works perfectly, though that app doesn't use all of the available features.

I'm still relatively new to Elixir, so any suggestions for cleaning up the code or making it more idiomatic are most welcome.

### DIFF
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -24,8 +24,8 @@ defmodule Bamboo.SendGridAdapter do
   """
 
   @service_name "SendGrid"
-  @default_base_uri "https://api.sendgrid.com/api"
-  @send_message_path "/mail.send.json"
+  @default_base_uri "https://sendgrid.com/v3/"
+  @send_message_path "/mail/send"
   @behaviour Bamboo.Adapter
 
   alias Bamboo.Email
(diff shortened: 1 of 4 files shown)


randycoulman:
I just rebased this on the latest master.

javierjulio:
@randycoulman thanks for doing this! I just started a small Elixir project last Thursday that we are now using in production. Its just meant to send 3 CSV reports as an email attachment using SendGrid. This has worked great for us! No issues sending emails to SendGrid with v3 API.

devshane:
Hi, any update on merging support for v3? I wonder if this is the cause of a mysterious "Unexpected error" mentioned in #280.

Schultzer:
@paulcsmith what is the status of this and #294?

randycoulman:
@paulcsmith Done

paulcsmith:
Thanks

PR #294, opened 2017-06-01
Support attachments in SendGridAdapter

randycoulman:
**WARNING** This PR is built using the v3 SendGrid API that is currently pending in #293, so that PR should be merged first and then this PR rebased on top of master.  All of the diffs that are unique to this PR start at commit dc8e927.

Adds attachment support to `SendGridAdapter`.  The code and tests are very similar to `MandrillAdapter`.

If #292 gets merged before this PR, I'm happy to make the necessary changes here.

### DIFF
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -167,6 +167,7 @@ defmodule Bamboo.SendGridAdapter do
   end
   defp put_template_substitutions(body, _), do: body
 
+  defp put_attachments(body, %Email{attachments: []}), do: body
   defp put_attachments(body, %Email{attachments: attachments}) do
     transformed = attachments
     |> Enum.reverse
@@ -179,7 +180,6 @@ defmodule Bamboo.SendGridAdapter do
     end)
     Map.put(body, :attachments, transformed)
   end
-  defp put_attachments(body, _), do: body
 
   defp ensure_content_provided(%{content: []} = body, email) do
     put_content(body, %Email{email | text_body: " "})
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -56,6 +56,9 @@ defmodule Bamboo.SendGridAdapter do
     end
   end
 
+  @doc false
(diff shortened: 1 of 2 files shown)


schurig:
@randycoulman #292 is merged  🎉  🌮

randycoulman:
@schurig I saw that.  My open source time is limited, but I'll try to get a rebase done when I can in the next day or two.

randycoulman:
I rebased #293 onto the latest master and then rebased this branch on top of that one.

randycoulman:
OK, changes made.  This one should be ready to go once #293 has been merged.

manukall:
hey, we've tried this branch and it works great when sending emails with attachment. it breaks for emails without attachment however. 
sendgrid requires the body to not include the `attachments` key at all in such cases. adding a guard clause `when attachments != []`  to `put_attachments` fixed the issue for us.

randycoulman:
@manukall Oh, good catch!  I thought I'd tested that in our application, but perhaps not.  I'll try to address that as soon as I can (probably in a day or two).

randycoulman:
@manukall I just pushed up another commit that addresses the issue you found.  I did it a bit differently: `Email`'s `attachments` field is always an array and not nil, so we don't need the final clause that I had written; instead, I've added an earlier clause that handles the empty array case.  Thanks for catching this!

javierjulio:
@randycoulman thanks for doing this! I just started a small Elixir project last Thursday that we've been using in production since then. Its just meant to send 3 CSV data dump reports as an email attachment using SendGrid. This has worked great for us! No issues sending emails with attachments from a local file or using raw data. Its worked flawlessly. Please let me know if you need help/confirmation on changes here as I can help test it as it gets reviewed. Thanks again! ❤️

randycoulman:
@paulcsmith I'll rebase once #293 has been merged.

paulcsmith:
#293 is in! Thanks so much for the quick response!

randycoulman:
@paulcsmith Rebased!

paulcsmith:
Thank you! :D

henry-hz:
great feature. let me suggest to add attachment to emails direct from a function call (instead of saving on the disk before sending).

PR #298, opened 2017-06-09
A backwards compatible link before v1.0 goes out

aforward:
I know that the preview module name has changed, but until v1.0 goes out, the link on the main page is referencing a hex doc that does not exist.  This simple PR is a stop gap to avoid the 404 page when visiting https://hexdocs.pm/bamboo/Bamboo.SentEmailViewerPlug.html

screen shot 2017-06-09 at 9 44 17 am

It now includes a pre v1.0 link

screen shot 2017-06-09 at 9 52 47 am


paulcsmith:
This is a good solution! Thank you so much!

PR #300, opened 2017-06-15
Added ability to set api key using env variables for SendGrid adapter

alekseenko:


maymillerricci:
It looks like this functionality was merged in in #342, so I'm going to close this. Thanks!

PR #301, opened 2017-06-17
Fix link to Sendgrid helpers hex docs in README

maxbeizer:
The link was a 404.

🍻


dbernheisel:
Thanks @maxbeizer, I just merged #315 into master which takes care of the broken link.

Issue #302, opened 2017-06-28
Unable to set header Reply-To through BambooPostmarkAdapter


technicalcapt:
So i am unable to put header `Reply-To` when using `BambooPostmarkAdapter`.
This is the error from Postmark api:
```
** (Bamboo.PostmarkAdapter.ApiError) There was a problem sending the email through the Postmark API.
Here is the response:
"{\"ErrorCode\":300,\"Message\":\"Header 'Reply-To' not allowed.\"}"
```
I am planning with a field `reply_to` in `Bamboo.Email`. Please take a look. Thank you!

PR #303, opened 2017-06-28
Bamboo.Email - support reply_to 

technicalcapt:
#302


samhamilton:
@paulcsmith - is there any else we can do to help get this merged?

technicalcapt:
@paulcsmith Thanks - Just try it again and it's working!. You can close #302

paulcsmith:
Awesome! I’m glad it’s working for you :)

markmark206:
i ran into a problem with the approach of 

```
...
|> put_header("reply-to", "foo@bar.com")
...
```

, when username / email address has special characters in it. 

In this case, the DKIM signature that my mailing service (Amazon SES) attaches to the email is deemed invalid by the receiving mailing service (i tested with gmail and hotmail).

For example, if my `reply-to` header looks like this: 
```
...
|> put_header("reply-to", "MarkMark 🍕🧃🎮 <redacted my gmail address@gmail.com>")
...
```

, gmail flags an invalid DKIM signature. I can see this In the "original email" view:

image

Removing special characters from the string fixes the problem. 

image

I am using Amazon SES with Bamboo.SMTPAdapter:

```elixir
  config :my_app, MyApp.Mailer.Bamboo,
    adapter: Bamboo.SMTPAdapter,
    server: "email-smtp.#{aws_region}.amazonaws.com",
...
```

Supplying the same user name / address as a tuple to the `from` field works just fine: 

```elixir
    new_email(
      from: {"MarkMark 🍕🧃🎮", "foo@example.com"},
      ...
```

A possible workaround is to simply use a bare email address in the `reply-to` header.

PR #304, opened 2017-07-01
Fixing compilation warnings

kelvinst:


lib/bamboo/phoenix.ex
```diff
@@ -1,287 +1,289 @@
...
-        def text_and_html_email_without_layouts do
-          new_email()
-          |> render(:text_and_html_email)
-        end
-
-        def email_with_assigns(user) do
-          new_email()
-          # @user will be available in the template
-          |> render(:email_with_assigns, user: user)
+if Code.ensure_loaded?(Phoenix) do
```
kelvinst: This is the only line added, the rest is only indentation.

kelvinst: BTW, there is another way to do it, removing `only: :test` from phoenix dep declaration. But I found this one a little bit better since bamboo itself does not depend on phoenix.

Going far away in my imagination, I guess it would be a good idea to separate some of Phoenix and Plug code from here. This could be an umbrella project with three applications: `bamboo` with the basics, `bamboo_phoenix` to use some phoenix views to render and `bambo_viewer` for the email viewer. 

Will be happy to open a PR to it if you liked the umbrella idea.

mravey:
Any change this get merged?

lib/bamboo/phoenix.ex
paulcsmith: Yes, I think at some point a `bamboo_phoenix` lib might be nice. For now this works great. Thank you so much!

kelvinst:
Thanks! Rebased!

paulcsmith:
Thanks a ton!

PR #306, opened 2017-07-04
Add support to Mailgun adapter for using custom vars

jfcalvo:
This PR add support for Mailgun custom variables as defined on [Mailgun documentation](https://documentation.mailgun.com/en/latest/user_manual.html#attaching-data-to-messages).

Tested and working with Mailgun.

Related issue #305

PR #307, opened 2017-07-13
Add attachment support for mailgun

schurig:
This PR adds **attachment support to the mailgun adapter**.

After #306 is merged this PR needs to be adjusted.

### DIFF
--- a/lib/bamboo/adapters/mailgun_adapter.ex
+++ b/lib/bamboo/adapters/mailgun_adapter.ex
@@ -20,24 +20,12 @@ defmodule Bamboo.MailgunAdapter do
   """
 
   @service_name "Mailgun"
-  @base_uri "https://api.mailgun.net/v3/"
+  @base_uri "https://api.mailgun.net/v3"
+  @api_endpoint "/messages"
   @behaviour Bamboo.Adapter
 
-  alias Bamboo.Email
+  alias Bamboo.{Email, Attachment}
   import Bamboo.ApiError
-
-  def deliver(email, config) do
-    body = email |> to_mailgun_body |> Plug.Conn.Query.encode
-
-    case :hackney.post(full_uri(config), headers(config), body, [:with_body]) do
-      {:ok, status, _headers, response} when status > 299 ->
-        raise_api_error(@service_name, response, body)
-      {:ok, status, headers, response} ->
-        %{status_code: status, headers: headers, body: response}
-      {:error, reason} ->
-        raise_api_error(inspect(reason))
-    end
-  end
 
   @doc false
   def handle_config(config) do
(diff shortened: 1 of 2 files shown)


schurig:
I haven't found a way to test the attachments for the mailgun adapter. Can anyone help me with it? 

It uses [multipart as content type](https://github.com/schurig/bamboo/blob/feature/mailgun-attachment-support/lib/bamboo/adapters/mailgun_adapter.ex#L158-L164) and I can't get the attachment by calling `assert_receive {:fake_mailgun, struct}` in the test.

rickclare:
@schurig Now that https://github.com/thoughtbot/bamboo/pull/306 has been merged, it is a simple matter to resolve the merge conflicts for this PR?  We'd really like to be able to start sending attachments via Mailgun.

a12e:
@schurig Hi, what is the status of this PR? Is any help needed to get it merged?

schurig:
@a12e @rickclare hi both of you. I would be happy to get help with this PR. I currently don't have time to work on it. Also I'm still stuck with the tests for the attachments https://github.com/thoughtbot/bamboo/pull/307#issuecomment-315197367

a12e:
After digging a little, here is what I found about the test issue.

This is the raw HTTP body that we send to Mailgun, which seems right:
```
-----------------------------mhcillixvhiobfhg
content-length: 12
content-type: application/octet-stream
content-disposition: form-data; name="from"

from@foo.com
-----------------------------mhcillixvhiobfhg
content-length: 15
content-type: application/octet-stream
content-disposition: form-data; name="h:X-My-Header"

my_header_value
-----------------------------mhcillixvhiobfhg
content-length: 9
content-type: application/octet-stream
content-disposition: form-data; name="html"

HTML BODY
-----------------------------mhcillixvhiobfhg
content-length: 10
content-type: application/octet-stream
content-disposition: form-data; name="subject"

My Subject
-----------------------------mhcillixvhiobfhg
content-length: 9
content-type: application/octet-stream
content-disposition: form-data; name="text"

TEXT BODY
-----------------------------mhcillixvhiobfhg
content-length: 0
content-type: application/octet-stream
content-disposition: form-data; name="to"


-----------------------------mhcillixvhiobfhg
content-length: 16
content-type: application/octet-stream
content-disposition: form-data; "name"="attachment"; "filename"="attachment.txt"

Test Attachment

-----------------------------mhcillixvhiobfhg--
```

This request is parsed by the `Plug.Parsers.MULTIPART` as configured in the fake server:
```elixir
    plug Plug.Parsers,
      parsers: [:urlencoded, :multipart, :json],
      pass: ["*/*"],
      json_decoder: Poison
```

This Parser should create a `:body_params` array in the Conn with all MIME fields. This is what we actually get in the Conn:
```
%Plug.Conn{...,
body_params: %{"bcc" => "BCC <bcc@bar.com>", "cc" => "CC <cc@bar.com>",
  "from" => "foo@bar.com", "subject" => "",
  "to" => "To <to@bar.com>,noname@bar.com"},
}
```

The `attachment` field has not been parsed. It does not appear in the Conn params.

The solution is to use Plug.Upload. I'll try to work on this. (I found it after writing this message)

schurig:
@a12e awesome thank you so much! Let me know how it goes. Also, feel free to open a new PR and use my changes if that works better for you!

schurig:
New PR: #341

PR #308, opened 2017-08-01
%Response{} struct to handle deliver responses

### LINKED ISSUE #161: Add message_id to Email struct definition
This may be a need unique to my situation, but it would help to be able to set a Message-ID prior to delivering an email so that the value can be saved to a database table and can be referenced later. Something similar is available in the [Ruby Tmail Library](https://github.com/mikel/tmail/blob/eb43e766cbad6125508f95223b8c0bc89cfdcff1/lib/tmail/interface.rb#L573) .

bnns:
This PR exposes the HTTP response from sending mail via the adapter.

- defines a `%Bamboo.Response{}` struct to store the responses.  Adapters now return a response with `Bamboo.Response.new_response/1`.
- defines a new `:response` key in `%Email{}` (can be `:nil` or `%Response{}`)
- merges the response from `adapter.deliver/2` with the email in `Mailer.deliver_now/3`
- leaves the `:response` key as `:nil` in the case we return `{:ok, pid}` from `Mailer.deliver_later/3`

Fixes #161

### DIFF
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -118,12 +118,15 @@ defmodule Bamboo.Mailer do
 
     if email.to == [] && email.cc == [] && email.bcc == [] do
       debug_unsent(email)
+      email
     else
       debug_sent(email, adapter)
-      adapter.deliver(email, config)
+      response =
+        email
+        |> adapter.deliver(config)
+        |> parse_response
+      email |> Map.merge(%{response: response})
     end
-
-    email
   end
 
   @doc false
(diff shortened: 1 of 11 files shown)


lib/bamboo/mailer.ex
```diff
@@ -131,13 +135,22 @@ defmodule Bamboo.Mailer do
...
+        |> parse_response
+      email |> Map.merge(%{response: response})
     end
-    email
   end
 
+  defp parse_response(%Bamboo.Response{} = response), do: response
+  defp parse_response({:ok, _pid}), do: nil
+  defp parse_response(:noop), do: nil
+  defp parse_response(response), do: raise "Unknown response from adapter: #{inspect response}"
```
bnns: Currently, a delivery strategy will need to return either `%Response{}`, `{:ok, pid}` or `:noop`. This might be too restrictive. Another option is to make the default case `nil` for anything besides `%Response{}` instead of raising.

Ch4s3:
It would be really nice to see this updated and merged.

Ch4s3:
What's the status of this PR? My team is currently running a fork, and we'd love to get right with the main repo. Let me know if I can help out in any way here.

Ch4s3:
Any update on when someone can take a look at this?

maymillerricci:
Thanks for your great work on this, and sorry for the delay! I opened #452 using parts of this as inspiration but not changing the type of what's returned from `deliver_now`. Let me know if you have any thoughts about it since you have great context for it!

bnns:
Closing since #452 seems to be a cleaner way to achieve the same thing.

PR #315, opened 2017-09-01
sendgrid helper link working

decentname:
Readme updated with correct link to sendgrid helper docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ vars, templates, and scheduling emails to send in the future. See
 
 SendGrid offers extra features on top of regular SMTP email like transactional
 templates with substitution tags. See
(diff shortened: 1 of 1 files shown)


dbernheisel:
@decentname  Thanks!

paulcsmith:
@dbernheisel I think this should be reverted or we need to make an issue to remind us to change it back before we release 1.0 because we renamed the module: https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/send_grid_helper.ex#L1

dbernheisel:
Thanks @paulcsmith. That's tough, because folks have run into problems because of the disparity between master and the published version. What do you think of editing the readme to point to both versions until 1.0?

paulcsmith:
@dbernheisel Yeah pointing to both sounds like a good idea to me

Issue #331, opened 2017-11-14
Allow sending custom args through SendGrid adapter


coryt:
It looks like there is no way to send custom args through to SendGrid.

I would like to add support for this and it looks like it could be done by using the `private` map and exposing a function within the SendGridHelper to add custom args...

Thoughts?

lancejjohnson:
@coryt If this is still an interest, please open a pull request with your recommended changes. Thanks!

germsvel:
Closing this issue (like we closed the associated PR https://github.com/thoughtbot/bamboo/pull/332) because it seems like https://github.com/thoughtbot/bamboo/pull/413 might be sufficient for sending custom args. If this should be reopened, please let me know. I'd be happy to do so.

PR #332, opened 2017-11-14
Adds support for sendgrid custom args

coryt:
This PR adds top level custom args support to the sendgrid adapter as mentioned in https://github.com/thoughtbot/bamboo/issues/331

### DIFF
--- a/lib/bamboo/adapters/send_grid_helper.ex
+++ b/lib/bamboo/adapters/send_grid_helper.ex
@@ -14,6 +14,24 @@ defmodule Bamboo.SendGridHelper do
 
   @id_size 36
   @field_name :send_grid_template
+  @custom_args :custom_args
+
+  @doc """
+  Specify a custom argument values that are specific to the entire send that will be carried along with the email and its activity data.
+  Substitutions are not made on custom arguments within SendGrid.
+  The combined total size of these custom arguments may not exceed 10,000 bytes.
+
+  ## Example
+
+      email
+      |> with_custom_args("userId", "12345")
+  """
+  def with_custom_args(email, arg_name, arg_value) do
+    custom_args = Map.get(email.private, @custom_args, %{})
+    |> Map.put(arg_name, arg_value)
+    email
+    |> Email.put_private(@custom_args, custom_args)
+  end
 
   @doc """
   Specify the template for SendGrid to use for the context of the substitution
--- a/lib/bamboo/adapters/send_grid_helper.ex
+++ b/lib/bamboo/adapters/send_grid_helper.ex
@@ -20,6 +20,7 @@ defmodule Bamboo.SendGridHelper do
   Specify a custom argument values that are specific to the entire send that will be carried along with the email and its activity data.
   Substitutions are not made on custom arguments within SendGrid.
   The combined total size of these custom arguments may not exceed 10,000 bytes.
+  Custom args must be a `String.t`.
 
   ## Example
 
@@ -27,10 +28,14 @@ defmodule Bamboo.SendGridHelper do
       |> with_custom_args("userId", "12345")
   """
   def with_custom_args(email, arg_name, arg_value) do
-    custom_args = Map.get(email.private, @custom_args, %{})
-    |> Map.put(arg_name, arg_value)
-    email
-    |> Email.put_private(@custom_args, custom_args)
+    if is_binary(arg_name) and is_binary(arg_value) do
+      custom_args = Map.get(email.private, @custom_args, %{})
+      |> Map.put(arg_name, arg_value)
+      email
+      |> Email.put_private(@custom_args, custom_args)
+    else
+      raise "expected the both arg_name and arg_value parameters to be of type binary, got #{arg_name}:#{arg_value}"
+    end
   end
 
   @doc """
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -92,6 +92,7 @@ defmodule Bamboo.SendGridAdapter do
     |> put_content(email)
     |> put_template_id(email)
     |> put_attachments(email)
+    |> put_custom_args(email)
   end
 
   defp put_from(body, %Email{from: from}) do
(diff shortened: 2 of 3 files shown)


doomspork:
This would be wonderful to have as part of the SendGrid support.  @paulcsmith what changes does this PR need to be considered?

lib/bamboo/adapters/send_grid_helper.ex
```diff
@@ -14,6 +14,29 @@ defmodule Bamboo.SendGridHelper do
...
+  The combined total size of these custom arguments may not exceed 10,000 bytes.
+  Custom args must be a `String.t`.
+
+  ## Example
+
+      email
+      |> with_custom_args("userId", "12345")
+  """
+  def with_custom_args(email, arg_name, arg_value) do
+    if is_binary(arg_name) and is_binary(arg_value) do
```
doomspork: @coryt do you think we need to enforce that the value is a string here or could we accept anything and call `to_string/1` below when we assign it?

The raise below is doing essentially that with `#{arg_name}:#{arg_value}`

doomspork:
@coryt as part of this PR should we support the unique arguments?
 https://sendgrid.com/docs/API_Reference/SMTP_API/unique_arguments.html

rockwood:
This would really help us out as well. We're trying to access the `send_at` argument, and it looks like this would solve it.

benjaminsfrank:
+1, would be great to get this merged in.

maymillerricci:
@coryt Thanks for the PR and sorry for the delay! Is this still desired behavior? If so, I think this will need to be updated to differentiate between top-level `custom_args` and personalization-level `custom_args` that were added in #413. Are you up for making the updates?

coryt:
> 
> 
> @coryt Thanks for the PR and sorry for the delay! Is this still desired behavior? If so, I think this will need to be updated to differentiate between top-level `custom_args` and personalization-level `custom_args` that were added in #413. Are you up for making the updates?

no problem, I should have some time to take another look

germsvel:
Hi @coryt, thanks for the patience with this PR. I'm taking over Bamboo maintenance and just got to it. 

I realize it's been a while since you first opened the PR, so if you're tired of it. Let me know. I can try to move it forward from here or ask someone else to take it over. But if you're still up for it, I left some thoughts below. Either way, thank you for your contribution! I learned a lot reading this PR and going through all the related docs. 

---

It seems like the ability to add custom args to personalizations was introduced in https://github.com/thoughtbot/bamboo/pull/413. With it, we can now do something like this `Email.put_private(email, :custom_args, %{post_code: "123"})`, and the SendGrid adapter will then set that map as `custom_args` on the personalizations. (See [personalizations section](https://github.com/thoughtbot/bamboo/blob/198b5042ff1521999ce9c51f4326ee5065dac650/lib/bamboo/adapters/send_grid_adapter.ex#L142) and [put_custom_args function](https://github.com/thoughtbot/bamboo/blob/198b5042ff1521999ce9c51f4326ee5065dac650/lib/bamboo/adapters/send_grid_adapter.ex#L286-L296).)

But #413 sets them as part of `personalizations`, and this PR is doing it directly on the `body`. It seems [SendGrid supports both](https://sendgrid.com/docs/for-developers/sending-email/personalizations/): 

> Some properties can be defined both at the root level of the request body (message level) and at the personalizations level.

> For example, the from, subject, headers, custom_arg, and send_at properties can all be defined at the message level or at the personalizations level. Individual fields within the personalizations array will override any message level properties that are defined outside of personalizations. For example, the email subject defined within personalizations will override the subject defined at the message level in the JSON payload.

> Keys within objects such as custom_args will be merged. If any of the keys conflict, the keys in the personalizations object will replace the message level object’s keys.

(and based on [the comment](https://github.com/thoughtbot/bamboo/pull/332/files#diff-01244714f120298359b81cecfa35de87ad857b7e3293e3ec0c16d9238667620fR20-R23) you added in the code, I imagine you already knew this) 

I wish we'd saved the `put_private(:custom_args, custom_args)` for the top level like you do in this PR and had something else for the personalizations (maybe through [add_personalizations](https://github.com/thoughtbot/bamboo/blob/198b5042ff1521999ce9c51f4326ee5065dac650/lib/bamboo/adapters/send_grid_helper.ex#L260-L264)). 

I don't know what's the best way to separate the two implementations so it's not confusing to bamboo users and so it's backwards compatible (the other `custom_args` were introduced in [v1.2](https://github.com/thoughtbot/bamboo/blob/master/CHANGELOG.md#120---2019-01-30).) Since SendGrid will handle the merging of the custom_args, I don't think that's something we should do within Bamboo itself. So I wonder if we'll have to use a _different key_ within `:private` for the message-level custom args, or if we should bundle this into a breaking change that changes how we do custom args for the personalization section. 

What do you think?

germsvel:
@coryt do you know if this PR is still something you want? I know there's a way to put `custom_args` at the personalizations level now, so maybe that is sufficient? 

If you still want this to go in, I'm planning on doing a v2 release for bamboo in the near future where we could include breaking changes. So we could sort out the clash between custom args at the personalizations level and at the root level in that release, if that's of interest. 

Thanks again for opening this PR.

lib/bamboo/adapters/send_grid_helper.ex
coryt: Sorry for the delay @doomspork 
This makes sense, I'll update this

coryt:
> @coryt do you know if this PR is still something you want? I know there's a way to put `custom_args` at the personalizations level now, so maybe that is sufficient?
> 
> If you still want this to go in, I'm planning on doing a v2 release for bamboo in the near future where we could include breaking changes. So we could sort out the clash between custom args at the personalizations level and at the root level in that release, if that's of interest.
> 
> Thanks again for opening this PR.

Thanks for the updates @germsvel and no problem, things have been busy on my end as well. 
I just took a look at the state of things and I think you are correct, it looks like https://github.com/thoughtbot/bamboo/pull/413 provides enough support so perhaps this isn't needed right now. 

Happy to close this PR for now.

germsvel:
Thanks for looping back and following on this issue @coryt! If you want to reintroduce this in the future, let me know. We can make it happen.

Issue #333, opened 2017-11-14
SendGrid error sending substitutions


coryt:
This is currently broken in master and was working before PR https://github.com/thoughtbot/bamboo/pull/293.

The issue seems to be that the payload being sent to SendGrid is using array of substitution values:
`"substitutions":{"test":["value"]}`
but the payload should be:
`"substitutions":{"text":"value"}`

This results in a cryptic response from the SendGrid api:
`"{\"errors\":[{\"message\":\"Bad Request\",\"field\":null,\"help\":null}]}"`

I have this fixed and working locally. I'll open a PR shortly.

manukall:
I'm running into the same problem. do you plan to merge the associated PR?

Strayer:
As I already commented in the PR, I think this is fixed with https://github.com/thoughtbot/bamboo/pull/337

PR #334, opened 2017-11-14
sendgrid substitution should send strings not array

coryt:
According to the documentation the substitutions should be `A collection of key/value pairs following the pattern "substitution_tag":"value to substitute". All are assumed to be strings.`

I was receiving a cryptic error response from the SendGrid api when using the latest code from Master.

More details in https://github.com/thoughtbot/bamboo/issues/333


Strayer:
I'm pretty sure this is fixed with https://github.com/thoughtbot/bamboo/pull/337

paulcsmith:
As @Strayer mentioned this should be fixed in master now :D

Issue #278, opened 2017-04-13
SendGridAdapter not available


dvanderbeek:
I'm getting an Internal Server Error when attempting to send via SendGrid on Heroku. Logs show this message:

```
** (UndefinedFunctionError) function Bamboo.SendGridAdapter.handle_config/1 is undefined (module Bamboo.SendGridAdapter is not available)
```

It does not error out when I try switching to Mailgun, but I would like to use SendGrid if possible.

paulcsmith:
@dvanderbeek Could you try using `Bamboo.SendgridAdapter` (note the lowercase g). The docs here on master are for the latest version of Bamboo. For the docs for the latest hex version, see these: https://hexdocs.pm/bamboo/0.8.0/readme.html

In there you will see that the adapter is spelled with a lower case `g`

LMK if that works for you :)

dvanderbeek:
@paulcsmith Yep, I just caught that and was about to update the issue here. Thanks for the quick reply!

paulcsmith:
You're welcome! I'm glad it's working now :)

antoniosb:
I'm getting this while trying to use `1.0.0-rc.1` with sendgrid:

```elixir
[error] beam/beam_load.c(1412): Error loading module 'Elixir.Bamboo.SendgridAdapter':
  module name in object code is Elixir.Bamboo.SendGridAdapter
```

happysalada:
this wrong spelling is still in the readme on master, people might still end up here.

dsignr:
^I ended up here because the readme has it wrong.

robacarp:
A little update. If your reading this, the spelling needs to be SendGrid now.

Upgrading our app from Bamboo ~0.8 to ~> 1.1.0, the spellinG changed from Sendgrid to SendGrid. 

We hit a similar error to the title of this issue and then dismissed it when "we had it right" because of the comment history (oops! 🤦‍♂️). At any rate, [as of a while ago](https://github.com/thoughtbot/bamboo/commit/6b582f80781f0072bd4051084a3286991bfde2d0) the correct spelling is [SendGrid](https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/send_grid_adapter.ex#L1)

mackshkatz:
@robacarp *you're :) Love you :)

kelraf:
Thanks, @paulcsmith your solution saved my day.

PR #336, opened 2017-11-19
Fix wrong spelling for Sendgrid adapter.

dsignr:
The Sendgrid adapter is incorrectly spelt with a "G" instead of "g" leading to https://github.com/thoughtbot/bamboo/issues/278.


paulcsmith:
SendGrid is spelled with a capital G. When 1.0 is fully tested and released the link will work. It's an unfortunate consequence that most people will see the README on master instead of checking on hex.pm I'll make sure to improve this process in the future though! I know it's not ideal right now

PR #337, opened 2017-11-22
Fix SendGrid templates for API v3

Strayer:
The SendGridAdapter forces content and subject for template-only SendGrid API requests. When using the recent master (we need attachments), HTML emails are completely broken. This only happens when having `{"content": [{"type": "text/plain", "value": " "}]}` in the API post request.

The error explained in the comment only happens when the transmitted content is an empty array - it doesn't occur if no content key appears at all.

I only ran the tests and did some testing with our SendGrid template setup, so maybe someone can also test non-template SendGrid requests (although I don't see why there should be any problem with those)

### DIFF
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -132,7 +132,12 @@ defmodule Bamboo.SendGridAdapter do
   defp put_subject(body, %Email{subject: subject}), do: Map.put(body, :subject, subject)
 
   defp put_content(body, email) do
-    Map.put(body, :content, content(email))
+    email_content = content(email)
+    if not Enum.empty?(email_content) do
+      Map.put(body, :content, content(email))
+    else
+      body
+    end
   end
 
   defp content(email) do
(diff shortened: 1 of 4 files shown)


Strayer:
Note: Still working on this, as there are some problems with the subject key.

Strayer:
This seems to work fine for us now. I had to remove the forced subject too, since it *must* be a string, if provided. When using templates the subject is not required.
Additionally for SendGrid API v3 requires substitutions to be a key=>value object where both should be strings. The current code sets the value as a list with one string. Couldn't find any info on SendGrid API v2, but I assume this was required for that version. Causes 400 Bad Request in v3 ;)

Strayer:
@paulcsmith Any chance to get this into the official tree?

Strayer:
@paulcsmith, sorry for being annoying… anything I should improve about this PR?

mschae:
:+1: @paulcsmith It would be great to have this upstream, we know other Bamboo users need this as well.

paulcsmith:
Thanks for the reminders and pings. Been a bit busy lately :D It's merged now. Thanks @Strayer!

PR #341, opened 2017-11-27
Mailgun attachment support

a12e:
Hi,

This is the followup of @schurig's great PR #307.

I fixed fixed a bug and merged his work into master.

The merging was bit tricky because the "custom variables" feature has been added during that time, generating nice merge conflicts. @schurig, I didn't understand why there are two identical `defp prepare_file` clauses in `lib/bamboo/adapters/mailgun_adapter.ex`.

I successfully tested with 1 attachment on the real Mailgun. (Should we test with several?)

Please let me know if you have any comment :)


lib/bamboo/adapters/mailgun_adapter.ex
```diff
@@ -116,17 +139,50 @@ defmodule Bamboo.MailgunAdapter do
...
-      # Key is a well known mailgun field or is an header or custom var field and its value is not empty
-      (key in @mailgun_message_fields || String.starts_with?(Atom.to_string(key), ["h:", "v:"])) && !(value in [nil, "", []])
+  defp prepare_file(%Attachment{path: nil} = attachment) do
+    {"", attachment.data,
+     {"form-data",
+      [{"name", ~s/"attachment"/},
+       {"filename", ~s/"#{attachment.filename}"/}]},
+     []}
+  end
+  # what is this second case???
```
schurig: Good catch, I think we can remove it. Or better to remove the first one that matches for `%Attachment{path: nil}` because we don't care about it's value

schurig:
@a12e LGTM :) I currently don't have a mailgun to test but maybe someone else has?

a12e:
Hi @schurig, I just want to 

- [ ]  Fix the misleading comments I added
- [ ]  Test with 2 attachments, and add a test case if it's problematic
- [ ]  Remove the extra `prepare_file` clause.

I'll try to do it this week. Then, it can definitely be merged :)

a12e:
So, finally it is not possible to test multiple attachments.

Mailgun wants multiple `attachment` fields (with the same name). But `Plug.Conn` cannot differentiate fields with the same name, because the the `params` of a `conn` are stored in a map, which don't allow multiple same keys.

However, I checked, multiple attachments are properly sent.

Everything is good to merge this?

schurig:
@a12e this only happens in the test environment right?

a12e:
Yes, this is only a test limitation.

The point is that we use `Plug.Conn` to test what we send to Mailgun (ie. parse the MIME body that we send into a `params` map.) And that `Plug.Conn`, by design, doesn't support params that have the same name (`attachement` for every attachment, as required by Mailgun)

schurig:
Great, thanks for clarifying. :)
From my side it's ready to go then! 🚀

pierot:
Any plans on merging this?

PR #348, opened 2018-01-25
Allow dynamically change mailer config

hauleth:
This adds `deliver_now/2` and `deliver_later/2` which allows user do dynamically change current mailer configuration. This is useful solution for cases when application need to dynamically change configuration (like multitenant application where user is allowed to send emails on own domain behalf).


adkron:
The functionality that is wanted in this PR is really useful for my current situation. Is this something, that if fixed, would be pulled in?

maymillerricci:
Closing in favor of #497.

PR #359, opened 2018-04-04
Allows dynamic configurations in deliveries

adkron:
At times we need custom configurations for sending specific emails. The
custom configurations might be needed because specific email addresses
need to move through specific servers. Instead of calling
`Bamboo.Mailer.deliver_*` directly and bypassing the public API this
changes the API to take in an optional second argument that is merged
into the default configuration. It will take in a map or a list of
keyword arguments.

Amos King @adkron <amos@binarynoggin.com>

### DIFF
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -2,7 +2,8 @@ defmodule Bamboo.Mailer do
   @moduledoc """
   Sets up mailers that make it easy to configure and swap adapters.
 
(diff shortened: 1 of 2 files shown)


adkron:
This is a working version of #348

adkron:
I created an adapter for this that solves this problem. After I use it in production, I will add in a PR to have it added to the README. For now, if anyone wants to try it, you can find the adapter at https://github.com/BinaryNoggin/bamboo_config_adapter and on hex.

PR #363, opened 2018-04-16
Allow to set mailgun config from env

Apelsinka223:
Ability to set config for mailgun adapter via env variables at runtime, by analogy with https://github.com/thoughtbot/bamboo/pull/342


redrabbit:
Would be great to have this PR merged. It's currently the standard way for deploying sensitive config data as part of Erlang releases.

hpopp:
Would love to see this merged 💖

PR #366, opened 2018-04-26
Remove warnings

JonRowe:
These lines were generating warnings when compiled, is it possible to get these cleaned up? I know that some of these may have been for git churn consistency but Elixir doesn't like them :/


paulcsmith:
Thanks!

Issue #329, opened 2017-11-14
Allow SendGrid Sandbox mode


MvanDiemen:
We are missing the sandbox mode for the sendgrid adapter. This is an extra block at the V3 request:

```
  "mail_settings": {
      "sandbox_mode": {
          "enable": true
      }
  }
```
We build this in and have a PR ready (with tests of course):
```
diff --git a/lib/bamboo/adapters/send_grid_adapter.ex b/lib/bamboo/adapters/send_grid_adapter.ex
index 49b21a8..537fcdc 100644
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -33,7 +33,7 @@ defmodule Bamboo.SendGridAdapter do
 
   def deliver(email, config) do
     api_key = get_key(config)
-    body = email |> to_sendgrid_body |> Poison.encode!
+    body = email |> to_sendgrid_body(config) |> Poison.encode!
     url = [base_uri(), @send_message_path]
 
     case :hackney.post(url, headers(api_key), body, [:with_body]) do
@@ -66,6 +66,13 @@ defmodule Bamboo.SendGridAdapter do
     end
   end
 
+  defp get_sandbox_mode(config) do
+    case Map.get(config, :sandbox_mode) do
+      nil -> false
+      key -> key
+    end
+  end
+
   defp raise_api_key_error(config) do
     raise ArgumentError, """
     There was no API key set for the SendGrid adapter.
@@ -83,7 +90,7 @@ defmodule Bamboo.SendGridAdapter do
     ]
   end
 
-  defp to_sendgrid_body(%Email{} = email) do
+  defp to_sendgrid_body(%Email{} = email, config = %{}) do
     %{}
     |> put_from(email)
     |> put_personalization(email)
@@ -92,6 +99,7 @@ defmodule Bamboo.SendGridAdapter do
     |> put_content(email)
     |> put_template_id(email)
     |> put_attachments(email)
+    |> put_mail_settings(config)
   end
 
   defp put_from(body, %Email{from: from}) do
@@ -110,6 +118,10 @@ defmodule Bamboo.SendGridAdapter do
     |> put_template_substitutions(email)
   end
 
+  defp put_mail_settings(body, config) do
+    Map.put(body, :mail_settings, %{sandbox_mode: %{enable: get_sandbox_mode(config)}})
+  end
```
And the tests:
```
[...]

PR #367, opened 2018-04-26
Allow sandbox mode for SendGrid adapter

JonRowe:
Allows the configuration of the sandbox mode on SendGrid, ala #329 .

Use case:

We're running a staging and production Heroku app with SendGrid and a mirrored DB snapshot, we'd like to set sandbox mode on staging so we can see emails have gone to the send grid, but not been sent.

### DIFF
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -16,6 +16,10 @@ defmodule Bamboo.SendGridAdapter do
       config :my_app, MyApp.Mailer,
         adapter: Bamboo.SendGridAdapter,
         api_key: "my_api_key" # or {:system, "SENDGRID_API_KEY"}
+        
+      # To enable sandbox mode (e.g. in development or staging environments),
+      # in config/dev.exs or config/prod.exs etc
+      config :my_app, MyApp.Mailer, sandbox: true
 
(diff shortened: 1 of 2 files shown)


JonRowe:
@paulcsmith done :)

JonRowe:
I cheated and used the web editor ;)

JonRowe:
Great! Thanks :) I know this is always a hassle but whats your release schedule like? Would love to get this into our codebase :)

ukutaht:
@JonRowe Sorry for bugging you on this.

Is feature working for you OK still? I have exactly the same use case as you do, but I can't get it to work. The emails are still going through and sent for real even when I specify `sandbox: true` in the configuration.

Could there be an issue with the request format? This line:
https://github.com/thoughtbot/bamboo/blob/907d761820096e8c3813676fee56e169300a2e0b/lib/bamboo/adapters/send_grid_adapter.ex#L163

Sets 
`{"mail_settings": {"sandbox": true}}`
whereas [SendGrid documentation](https://sendgrid.com/docs/for-developers/sending-email/sandbox-mode/) says it should be
 `{"mail_settings": {"sandbox_mode": {"enable": true}}`

Ideas?

JonRowe:
@ukutaht Sorry I no longer have access to the project using this, so I can't check, you are right about the documentation though!

PR #379, opened 2018-05-01
Unassign unused variables

drapergeek:
This removes the compiler warnings.

I would love to find a way to get these warnings to break the build but unfortunately `Phoenix.View` is used in our modules but is not in the mix file (for good reason). This means that we can't run `mix compile --warnings-as-errors` successfully.

Once we move to Elixir 1.6, I believe the autoformatter will take care of those issues. Any ideas in the meantime?


paulcsmith:
Oops, I missed this so this was closed by! https://github.com/thoughtbot/bamboo/commit/59ec8fa1cf305ac8fd0ed01bde6f37a639ca9010

I think in the meantime we can keep it as-is and wait to upgrade to Elixir 1.6

PR #386, opened 2018-05-09
Allow protocol impls to return a String.t

evuez:
The spec for `Bamboo.Formatter` says `format_email_address` should
return a `Bamboo.Email.address`, defined as `String.t | {String.t, String.t}`,
but returning a `String.t` didn't work with any of the adapters.

PS: There's some diff garbage because some whitespaces were removed, so it's easier to read with these ignored: https://github.com/thoughtbot/bamboo/pull/386/files?utf8=%E2%9C%93&diff=split&w=1 -- sorry about that.


germsvel:
I opened a PR to fix the typespec instead of supporting the changes, since it seems like the typespec allowed `String.t()` accidentally. Thanks for catching this, and opening a PR. Let me know if PR https://github.com/thoughtbot/bamboo/pull/570 seems incorrect to you.

evuez:
@germsvel #570 looks good to me :relaxed:

Issue #355, opened 2018-03-14
deliver/2 shouldn't raise exception


chingan-tsc:
Hi I am currently using a Postmark adapter for Bamboo and I am noticing that `deliver_now/1` will raise an exception if something goes wrong, now AFAIC it might be confusing as functions that raise exception should be appended with a bang !. 

I am proposing maybe we should have both `deliver_now!` and `deliver_now` in the spec that all the adapters can then respect the spec later on. What do you think?

Related issue: https://github.com/pablo-co/bamboo_postmark/issues/20

poops:
I came here to start the same discussion. Currently I have a script that bulk sends, and a timeout is throwing an exception and blocking the rest from sending. Wasn't expecting `deliver_now/1` to raise.

awochna:
I agree. Additionally, it's probably worth noting the Bamboo.TestAdapter doesn't have a clear way to help you test how your application will handle an exception being raised during delivery.

codyjroberts:
Agreed. The option to have standard `{:ok, _}/{:error, _}` return values would be handy.

PR #393, opened 2018-05-30
Don't raise when calling deliver/2 or deliver_later/2

poops:
See https://github.com/thoughtbot/bamboo/issues/355

### DIFF
--- a/lib/bamboo/adapters/mailgun_adapter.ex
+++ b/lib/bamboo/adapters/mailgun_adapter.ex
@@ -19,12 +19,10 @@ defmodule Bamboo.MailgunAdapter do
       end
   """
 
-  @service_name "Mailgun"
   @base_uri "https://api.mailgun.net/v3"
   @behaviour Bamboo.Adapter
 
   alias Bamboo.{Email, Attachment}
-  import Bamboo.ApiError
 
   @doc false
   def handle_config(config) do
@@ -51,14 +49,11 @@ defmodule Bamboo.MailgunAdapter do
     body = to_mailgun_body(email)
 
     case :hackney.post(full_uri(config), headers(email, config), body, [:with_body]) do
-      {:ok, status, _headers, response} when status > 299 ->
-        raise_api_error(@service_name, response, body)
-
       {:ok, status, headers, response} ->
-        %{status_code: status, headers: headers, body: response}
+        {:ok, %{status_code: status, headers: headers, body: response}}
 
-      {:error, reason} ->
-        raise_api_error(inspect(reason))
+      resp ->
+        resp
     end
   end
 
--- a/lib/bamboo/adapters/mandrill_adapter.ex
+++ b/lib/bamboo/adapters/mandrill_adapter.ex
@@ -19,13 +19,10 @@ defmodule Bamboo.MandrillAdapter do
       end
   """
 
-  @service_name "Mandrill"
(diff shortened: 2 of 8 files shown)


joshchernoff:
wanna try and resolve the merge?

poops:
Sorry accidentally closed this with a force push, but should be updated!

germsvel:
Hi @poops, thanks for opening this PR! I'm starting to maintain Bamboo, so it's taking a little while to catch up with all the issues/PRs. 

I know this is a breaking change, so I want to hold off until we're ready for a v2. Or rather, I just want to make sure to bundle all the breaking changes into one release so we don't have to release a few major versions in a sequence. 

So hold off on this for me. I'll loop back to this PR when we're ready.

germsvel:
Hi @poops, thanks for the patience with this PR. I think we can start moving forward with these changes. 

Ideally, I'd like to stop raising when we call `deliver/2` and `deliver_later/2` (what you did in this PR), but I also want to keep the existing behavior under `!` versions of those functions. 

That way people who don't care about the return value have an easy path for fixing breaking changes: they change `deliver` -> `deliver!` and `deliver_later` -> `deliver_later!`. 

So ideally we have something like this (which I mentioned in a related issue https://github.com/thoughtbot/bamboo/issues/505#issuecomment-690287834): 

```elixir
# success deliver_now
{:ok, email} = deliver_now(email)
{:ok, email, response} = deliver_now(email, response: true)

# error deliver_now
{:error, email} = deliver_now(email)
{:error, email, response} = deliver_now(email, response: true)

# success deliver_now!
email = deliver_now!(email)
{email, response} = deliver_now!(email, response: true)

# error deliver_now!
deliver_now!(email) # raises
deliver_now!(email, response: true) #raises
```

I've also been thinking about releasing a version before 2.0 that basically raises deprecation warnings and allows people to opt-in to the new behavior via configuration before 2.0. I'm trying to be extra careful because I wouldn't want someone to update to 2.0, _not_ change `deliver` -> `deliver!` and think everything is working as usual, but their emails are now failing silently because we no longer raise on errors. 

I realize those changes are outside of the scope of what you were trying to do here. So, I can go ahead and open a separate PR with all those changes. I can use what you have here as a starting point and set you as a co-author. Or if you want to go through all those layers, I'm happy to review PRs as well.

poops:
Hi @germsvel 

The deprecation warnings sound like a great idea. Since it sounds like you're doing that piece separately, I could update this PR once that's merged. Or if you want to start a new PR feel free to close this. No need to set me as co-author, just happy to have a way to call this w/o raising. 😄

germsvel:
@poops I couldn't find a good way to add an in-between release that would do deprecations warnings. But.. I managed to make it so that `deliver_now!` and `deliver_later!` have the same behavior that the current `deliver_now` and `deliver_later` have. So the upgrade path for those who aren't interested in handling errors would be to simply change: 

- `deliver_now` -> `deliver_now!` and
- `deliver_later` -> `deliver_later!`. 

I have opened a PR against the Bamboo 2 branch with this work. https://github.com/thoughtbot/bamboo/pull/571 I'd love your thoughts if you're interested in looking at it. 

Also, I think #571 supersedes this PR, so we can probably close it, if that sounds good to you.

poops:
Looks great, thanks for your work on this!

PR #401, opened 2018-06-14
Bump Hackney to 1.12.1

bratsche:
Bamboo is currently using a pretty old Hackney so that it's incompatible with `stripity_stripe`, because that package requires 1.12.1. So I wanted to see if you would be willing to bump it up to the same version in Bamboo.


paulcsmith:
👍 Thanks for the PR!

PR #410, opened 2018-07-05
Add supports_attachments? to adapter behaviour

kalys:


kalys:
```apt-get update``` failed on circleci

**Comment by joshchernoff** [comment]
Looks like maybe the docker image is having issue. 
```
W: GPG error: https://apt.dockerproject.org ubuntu-precise Release: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY F76221572C52609D
```

joshchernoff:
Oh wait it was 404. I would try just re running it. I but their server was just down.

kalys:
@paulcsmith @joshchernoff Do you think if we need these changes? Should ```supports_attachments?``` be part of API?

kalys:
@maymillerricci rebased

maymillerricci:
Perfect, thank you!

PR #411, opened 2018-07-06
Migrate circleci configuration to 2.0

kalys:


joshchernoff:
Why was there a change made to mix.lock ?

.circleci/config.yml
```diff
@@ -0,0 +1,44 @@
...
+      - save_cache:
+          key: v1-build-cache-{{ .Branch }}
+          paths: "_build"
+      - save_cache:
+          key: v1-build-cache
+          paths: "_build"
+      - run: mix test
+      - run: MIX_ENV=prod mix compile
+
+      - store_test_results:  # upload test results for display in Test Summary
```
joshchernoff: Not sure storing test results is necessary, though I could be wrong.

.circleci/config.yml
```diff
@@ -0,0 +1,44 @@
...
+          key: v1-mix-cache
+          paths: "deps"
+      - save_cache:
+          key: v1-build-cache-{{ .Branch }}
+          paths: "_build"
+      - save_cache:
+          key: v1-build-cache
+          paths: "_build"
+      - run: mix test
+      - run: MIX_ENV=prod mix compile
```
joshchernoff: Needed?

.circleci/config.yml
```diff
@@ -0,0 +1,44 @@
...
+      - run: mix do deps.get, compile
+      - save_cache:
+          key: v1-mix-cache-{{ .Branch }}-{{ checksum "mix.lock" }}
+          paths: "deps"
+      - save_cache:
+          key: v1-mix-cache-{{ .Branch }}
+          paths: "deps"
+      - save_cache:
+          key: v1-mix-cache
+          paths: "deps"
```
paulcsmith: I haven't used Circle 2.0 that much yet. Could you explain why multiple caches are needed here?

.circleci/config.yml
paulcsmith: I've found this is helpful to catch things that may only cause issues in prod.

kalys:
@joshchernoff @paulcsmith 
I used this file https://circleci.com/docs/2.0/language-elixir/#sample-configuration

Maybe cost of network traffic is higher than cost of storage and that's why they recommend to cache every version, branch and change of mix.lock.

kalys:
@joshchernoff @paulcsmith Ok. I removed all caches except one that depends on checksum of mix.lock. Also I removed ```v1-``` prefix from cache name.

paulcsmith:
Thank you so much!

PR #418, opened 2018-07-24
Allow to use the complete API URL for the Mailgun adapter

milmazz:
This commit allows to use the complete API URL instead of your domain for the
Mailgun adapter.

When I was trying to use this adapter for the first time I tried to follow the
same pattern as in `curl`:

```
curl -s --user 'api:key-3ax6xnjp29jd6fds4gc373sgvjxteol0' \
https://api.mailgun.net/v3/samples.mailgun.org/messages \
 -F from='Excited User <excited@samples.mailgun.org>' \
 -F to='devs@mailgun.net' \
 -F subject='Hello' \
 -F text='Testing some Mailgun awesomeness!'
```

It was a mistake trying to do this, but the current implementation do not emit
any warning about this. So, instead of emit a warning I added support for
passing the complete API URL instead of the domain.

### DIFF
--- a/lib/bamboo/adapters/mailgun_adapter.ex
+++ b/lib/bamboo/adapters/mailgun_adapter.ex
@@ -50,7 +50,7 @@ defmodule Bamboo.MailgunAdapter do
   def deliver(email, config) do
     body = to_mailgun_body(email)
 
-    case :hackney.post(full_uri(config), headers(email, config), body, [:with_body]) do
+    case :hackney.post(full_uri(config.domain), headers(email, config), body, [:with_body]) do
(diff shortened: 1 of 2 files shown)


joshchernoff:
Could you explain why you would want to provide the API Base URL vs just the Domain? 

I guess I fail to see why this is useful if you can provide the domain and it works why would you want to provide the full API Base URL? Since the config asks for the domain which is different then the API Base URL I also don't like how they get conflated.

milmazz:
@joshchernoff Yes, the `domain` is enough for the configuration and it works as is.

Issue #420, opened 2018-08-04
Bamboo.MailgunAdapter.put_reply_to/2 blows up because its setting a string and not an atom


joshchernoff:
Setting `"h:Reply-To",`  via 

```
  defp put_reply_to(body, %Email{headers: %{"reply-to" => address}}),
    do: Map.put(body, "h:Reply-To", address)
```
https://github.com/thoughtbot/bamboo/blob/6f10c5887a58ec7c3ace8b1fa8fa6fb3d9bf5c1a/lib/bamboo/adapters/mailgun_adapter.ex#L105

this will blow up blowing up because `"h:Reply-To"` gets passed set as a string in body, the when Enum.filter runs key (ie `"h:Reply-To"`) is what  is passed to Atom.to_string(key) as seen here
```
def filter_non_empty_mailgun_fields(body) do
    Enum.filter(body, fn {key, value} ->
      # Key is a well known mailgun field (including header and custom var field) and its value is not empty
      (key in @mailgun_message_fields || key in @internal_fields ||
         String.starts_with?(Atom.to_string(key), ["h:", "v:"])) && !(value in [nil, "", []])
    end)
    |> Enum.into(%{})
  end
```
https://github.com/thoughtbot/bamboo/blob/b0c66e5880ec970f5d46757a64bbebb14962659f/lib/bamboo/adapters/mailgun_adapter.ex#L171

You will end up with a:
```
** (ArgumentError) argument error
    :erlang.atom_to_binary("h:Reply-To", :utf8)
    (elixir) lib/enum.ex:900: anonymous fn/3 in Enum.filter/2
    (stdlib) lists.erl:1263: :lists.foldl/3
    (elixir) lib/enum.ex:1941: Enum.filter/2
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
    (iex) lib/iex/evaluator.ex:249: IEx.Evaluator.handle_eval/5
    (iex) lib/iex/evaluator.ex:229: IEx.Evaluator.do_eval/3
    (iex) lib/iex/evaluator.ex:207: IEx.Evaluator.eval/3
    (iex) lib/iex/evaluator.ex:94: IEx.Evaluator.loop/1
    (iex) lib/iex/evaluator.ex:24: IEx.Evaluator.init/4
    (iex) lib/iex/pry.ex:64: IEx.Pry.pry/2
[...]

PR #421, opened 2018-08-04
fix reply-to header being set as string for mailgun adpt

joshchernoff:
relates to #420 where h:Reply-To is being set in the body as a string vs an atom. This blows up a little down stream when Atom.to_string is called and passed the string of "h:Reply-To"


joshchernoff:
I also added a a fix to circle ci since you are still using v1 I bumped it to using v2

circle.yml
```diff
@@ -1,14 +1,44 @@
...
+          paths: "deps"
+      - save_cache: # don't forget to save a *build* cache, too
+          key: v1-mix-cache-{{ checksum "mix.lock" }}
+          paths: "_build"
+      - save_cache: # don't forget to save a *build* cache, too
+          key: v1-build-cache-{{ .Branch }}
+          paths: "_build"
+      - save_cache: # and one more build cache for good measure
+          key: v1-build-cache
+          paths: "_build"
```
paulcsmith: Could you explain all these caches though before I merge in? I'm not sure why another cache is needed for _build and deps

joshchernoff: To be honest I'm not 100% sure you even need that. I was just going off their site at 
https://circleci.com/docs/2.0/language-elixir/

That being said I see you merged the other PR for circle ci I can just drop that from this.

joshchernoff: oh wait, sorry I thought you had. 

Well, from what I can read. It looks like its a base cache for build vs the other { .Branch } one. 

I have used this config on other projects and it does not look to be an issue running in terms of resources. Though I really can't say if you will see less or more performance dropping the caches.

paulcsmith:
The Circle 2.0 PR has been merged. Could you rebase this and push it up without your Circle changes?

joshchernoff:
Done.

joshchernoff:
ping...

joshchernoff:
Ping Pong

joshchernoff:
@paulcsmith I did a rebase the commit #9786cf7 removed the config.yml and added that .circleci/circle.yml from upstream. 

Maybe I'm missing something?

joshchernoff:
@paulcsmith let me know if there was anything else you wanted to fix.

PR #431, opened 2018-09-09
Aallow for config driven base_uri so you can set MailgunAdapter to EU API endpoints

pierot:


pierot:
What do I need to do, to have this merged?
thanks

foeken:
Ping

kannix:
Another ping.. i also need this 😸

maymillerricci:
@pierot, @foeken, @kannix Sorry for the delay! I believe the desired functionality can actually already be achieved. It's already set up to fetch the base URI from the environment and use the default only if none is configured - see https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/mailgun_adapter.ex#L69.
If you add the EU-specific base URI to your environment config does it work like you want? 
```
config :bamboo, mailgun_base_uri: "base_uri_to_override"
```
Can you verify whether that works?

I see that this is not actually documented anywhere which we should resolve. 😄

kannix:
> Can you verify whether that works?

Yip, adding `config :bamboo, mailgun_base_uri: "https://api.eu.mailgun.net/v3"` to the config works 👍

maymillerricci:
Awesome, thanks for confirming! I'm going to close this since adding the base URI to the environment config works for this.

RichGuk:
@maymillerricci It might be a good idea to add this to the documentation somewhere? I had to hunt around in the code to find that this could be set.

arjan:
This is still an issue for multi-tenant setups where some tenants use EU domains and some tenants use US domains. Why wasn't a6202443e1438e5fb9913b7441623c087ac4a01f merged? Seems like a better approach (except for the Application env variable overriding the one from config).

Issue #409, opened 2018-06-25
Return values and exceptions for `Bamboo.Mailer.html.deliver_now/1`


nathanl:
I'm using Bamboo with https://github.com/fewlinesco/bamboo_smtp.

[The docs for `Bamboo.Mailer.deliver_now/1`](https://hexdocs.pm/bamboo/Bamboo.Mailer.html#deliver_now/1) don't say what return value to expect in case of success or what return value or exception I should expect in case of failure.

My experiments say that on success, I'll get back a bare `%Bamboo.Email{}` (not wrapped in `{:ok, val}`. If I use a bogus SMTP server name or credentials, a `%Bamboo.SMTPAdapter.SMTPError{}` is raised.

Sending to a bogus recipient returns a `%Bamboo.Email{}` as if it were successful, presumably because the SMTP server responded successfully and only later realized it couldn't deliver the email.

I'm not sure whether or how you'd want this documented, but I figured that someone might at least stumble on this issue and be helped a bit.

stefmona:
Would like to see something like that for Mandrill. 👍

germsvel:
I'm gonna go ahead and close this since Bamboo can now return ok/error tuples since https://github.com/thoughtbot/bamboo/pull/571. Let me know if you think this should still be open. Happy to reopen it if that's the case.

PR #450, opened 2019-02-11
Issue 409 return value from adapter

ckoch-ncsa:
Addresses [#409]. 

Returns the value from the Adapter rather than the Email struct. This can be important in the case of transactional email senders (for example Mandrill and Sendgrid) where you want/need to store the id generated by the underlying adapter service.

### DIFF
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -20,7 +20,7 @@ defmodule Bamboo.Mailer do
 
       # Somewhere in your application. Maybe lib/my_app/mailer.ex
       defmodule MyApp.Mailer do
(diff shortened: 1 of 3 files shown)


maymillerricci:
Thanks for your great work on this! I opened #452 using parts of this as inspiration but not introducing a new function for this. Let me know if you have any thoughts about it since you have great context!

ckoch-ncsa:
@maymillerricci 
Thank you for running with my PR. #452 looks like a great improvement. Should I close out this PR?

maymillerricci:
@ckoch-ncsa Yes, I think this can be closed. Thanks for your work on it!

PR #452, opened 2019-02-21
Conditionally include the adapter response in deliver_now

### LINKED ISSUE #451: Response values from adapter (deliver_now) behavior
Related to issue [#409]

When sending a message via some of the adapters, certainly Mandrill and SendGrid (possibly others), the email service API returns a unique message id as well as other service related metadata. Often these services will return some kind of message received status and return a list of errors in the case where something didn't go right. 
This message_id and other metadata from the service is useful and desirable by the implementations of the Bamboo mailer adapter. 

Would it be possible to return any service response in the :deliver_now function or would it be preferable to create a new function that returns the response from the adapter?

maymillerricci:
Resolves #451 and #161.

The email adapters return data about the results from sending the email when `.deliver` is called, some of which can be valuable but is not currently being captured. 

This introduces a change to merge that response in to the `Email` struct that is returned from `.deliver_now` if the response conforms to our norm of returning a status_code, headers, and body. (See the return for the success case from the adapters' `deliver` function for [SendGrid](https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/send_grid_adapter.ex#L49), [Mandrill](https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/mandrill_adapter.ex#L41), and [Mailgun](https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/mailgun_adapter.ex#L58).) If the return value does not conform to that pattern, nothing additional happens, so for example the results from `.deliver_now ` for the `LocalAdapter` and `TestAdapter` do not change. 

In looking at the custom adapters, it looks like some of them conform to this pattern, so they would get the response included now, and others do not and some have separate custom handling for their responses, so there would be no change for them. 

My goal was to include this data but to change as little as possible. I took inspiration from #308 and #450, but was opting not to change the type of what is returned from `deliver` or introduce a separate function for this.

### DIFF
--- a/lib/bamboo/email.ex
+++ b/lib/bamboo/email.ex
@@ -75,7 +75,8 @@ defmodule Bamboo.Email do
           text_body: nil | String.t(),
           headers: %{String.t() => String.t()},
           assigns: %{atom => any},
-          private: %{atom => any}
+          private: %{atom => any},
+          response: nil | %{atom => any}
         }
 
   defstruct from: nil,
@@ -88,7 +89,8 @@ defmodule Bamboo.Email do
             headers: %{},
             attachments: [],
             assigns: %{},
-            private: %{}
+            private: %{},
+            response: nil
 
   alias Bamboo.{Email, Attachment}
 
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -118,12 +118,12 @@ defmodule Bamboo.Mailer do
 
     if email.to == [] && email.cc == [] && email.bcc == [] do
       debug_unsent(email)
+      email
     else
       debug_sent(email, adapter)
-      adapter.deliver(email, config)
+      response = adapter.deliver(email, config)
+      maybe_merge_response(email, response)
     end
-
-    email
   end
 
   @doc false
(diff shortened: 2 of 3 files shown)


maymillerricci:
Upon reflection, I'm on the fence about this approach. It does solve the current issue without introducing any breaking changes which is great. But, it feels like `response` doesn't belong as part of `Email` which is meant to be all about creating/composing the email. But, I also don't love the other potential solutions, so I'm not sure what the best path forward with this is...

Ch4s3:
I think it would make sense to go with this approach and make cleaner but breaking changes as part of a major version bump.

maymillerricci:
@lancejjohnson Interesting! Thanks for your thoughts on this. I was initially resistant to the idea of creating a separate function for this, but that was for a function with a totally different name, and I hadn't considered the idea of adding something like a `response: true` option to the existing function. That might be a really good solution, at least for the short term. 

Ultimately doing a major version bump with breaking changes might make sense, and following the `{:ok, thing} | {:error, thing}` pattern seems like a natural fit. 

But for now, I like your idea, and it does feel better than merging the `response` onto the `Email` where it doesn't totally belong. The `Plug.Conn` parallel is interesting, but like you said in our case it feels different since something has left the system and come back, and also one of the explicit goals here seems to be for email creation and email delivery to be totally separate, so as much as I was wanting this to be the solution, it doesn't seem quite right. 

I will take a pass at implementing your suggestion of having a separate function with a `response` option, and we can see what we think of it. Thanks! 🙏

maymillerricci:
@lancejjohnson See #466 for my pass at your suggestion! Is something like that what you had in mind?

maymillerricci:
Closing this in favor of #466.

Issue #449, opened 2019-01-29
Key should be "filename" instead of "filname"


thadwoodman:
https://github.com/thoughtbot/bamboo/blob/23b8dc85d1cb095448e0f596759166fcc2caaf0c/lib/bamboo/email.ex#L217

maymillerricci:
Thanks for reporting! #454 fixes this.

PR #454, opened 2019-02-26
Fix typos and documentation

maymillerricci:
Minor fixes including typos and updates to code examples in the README that came up as I was trying to follow along setting up Bamboo. 

Includes a typo fix for https://github.com/thoughtbot/bamboo/issues/449.


README.md
```diff
@@ -241,13 +241,13 @@ You can use the Bamboo.TestAdapter along with [Bamboo.Test] to make testing your
 emails straightforward.
 
 ```elixir
-# Using the mailer from the Getting Started section
```
maymillerricci: This example really assumes more than the mailer from the Getting Started section, so I think it's better just not to say anything about it.

PR #456, opened 2019-02-27
Add API endpoints to SentEmailViewerPlug

tomtaylor:
Added two endpoints to the `SentEmailViewerPlug` to list and reset the sent emails programatically. These are useful for integration test runners, so it's possible to write tests that can check and respond to emailed content.

### DIFF
--- a/lib/bamboo/plug/sent_email_api/sent_email_api_plug.ex
+++ b/lib/bamboo/plug/sent_email_api/sent_email_api_plug.ex
@@ -1,0 +1,111 @@
+defmodule Bamboo.SentEmailApiPlug do
+  use Plug.Router
+  alias Bamboo.SentEmail
+
+  @moduledoc """
+  A plug that can be used in your router to expose delivered emails over a
+  JSON API. This is useful for an integration test runner that doesn't have
+  access to the `Bamboo.Test` helpers, but needs to be able to assert that
+  emails have been delivered.
+
+  You must use the `Bamboo.LocalAdapter`, or no emails will be available.
+
+  ## Using with Plug or Phoenix
+
+      # Make sure you are using Bamboo.LocalAdapter in your config
+      config :my_app, MyApp.Mailer,
+        adapter: Bamboo.LocalAdapter
+
+      # In your Router
+      defmodule MyApp.Router do
+        use Phoenix.Router # or use Plug.Router if you're not using Phoenix
+
+        if Mix.env == :e2e do
+          # If using Phoenix
+          forward "/sent_emails_api", Bamboo.SentEmailApiPlug
+
+          # If using Plug.Router, make sure to add the `to`
+          forward "/sent_emails_api", to: Bamboo.SentEmailApiPlug
+        end
+      end
+
+  ## API endpoints
+
+  There are two endpoints:
+
+  * `GET /emails.json` exposes the list of sent emails
+  * `POST /reset.json` resets the sent emails back to an empty list
+
+  Usually, you would `POST` to `/reset.json` at the beginning of a test run,
+  and `GET` from `/emails.json` as you need to during an integration test to
+  make assertions. The details of how you make assertions is left up to your
+  integration test runner.
+
+  ## Email response JSON
+
+  Sent emails are returned in the following format:
+
+  ```json
+  {
+    "to": [[null, "to@example.com"], ["Alice", "alice@example.com"]],
+    "text_body": "hello world",
+    "subject": "This is a test email",
+    "html_body": "<p>hello world</p>",
+    "headers": {},
+    "from": [null, "from@example.com"],
+    "cc": [[null, "cc@example.com"]],
+    "bcc": [["Bob", "bob@example.com"]],
+    "attachments": []
+  }
+  ```
+
+  Email addresses are returned in array pairs, with the first element as the
+  name, and the second as the email address. The first element can be `null`
+  if the address didn't specify a name. The `to`, `cc` and `bcc` fields are
+  returned as arrays of these pairs.
+  """
+
+  plug(:match)
+  plug(:dispatch)
+
+  get "/emails.json" do
+    data =
+      SentEmail.all()
+      |> Enum.map(&format_json_email/1)
+
+    conn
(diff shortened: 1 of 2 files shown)


maymillerricci:
@tomtaylor Interesting! Can you share an example of how you are thinking to use this? Is this for integration testing email sending? Would it be possible to achieve this using helpers from [Bamboo.Test](https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/test.ex)?

tomtaylor:
That’s right. I use [Cypress](https://www.cypress.io/) for our E2E tests, which can only interact with my Phoenix app over HTTP routes. I expose a couple of test only controller actions for setting up database scenarios, but had no good way of testing journeys with an email step in the middle. With this I can write a function that tell Cypress to reach into the Bamboo sent emails API to assert an email was received.

This isn’t happening in process in Elixir, so the `Bamboo.Test` helpers don’t help here. Does that make sense?

maymillerricci:
Ah, okay, that makes sense. What would you think about making these API endpoints in a separate plug that people could then expose in their apps only if they want to? I feel like the use case for these might be different enough to warrant it being separate, and many people would not need these exposed, although I'm not sure and would love to hear your thoughts. In any case it'd be great to have some documentation on here about what these are for. Thanks for your work on this!

tomtaylor:
Thanks @maymillerricci. Happy to do that if you feel it's clearer - I think it's likely that they wouldn't both be used at the same time, so that makes sense. And I'll add some documentation at the same time.

test/lib/bamboo/plug/sent_email_viewer_plug_test.exs
```diff
@@ -190,6 +190,44 @@ defmodule Bamboo.SentEmailViewerPlugTest do
     assert conn.resp_body =~ "Email not found"
   end
 
+  test "list emails over API" do
+    normalize_and_push_pair(:html_email,
```
maymillerricci: Thoughts on using `normalize_and_push` to create a single email since the assertions are all about the first one? Or alternatively doing assertions on both of the emails?

test/lib/bamboo/plug/sent_email_viewer_plug_test.exs
```diff
@@ -190,6 +190,44 @@ defmodule Bamboo.SentEmailViewerPlugTest do
...
+    assert {"content-type", "application/json; charset=utf-8"} in conn.resp_headers
+
+    json = Bamboo.json_library().decode!(conn.resp_body)
+
+    first_email = json |> Enum.at(0)
+    assert first_email["from"] == [nil, "from@example.com"]
+    assert first_email["to"] == [[nil, "to@example.com"], ["Alice", "alice@example.com"]]
+    assert first_email["cc"] == [[nil, "cc@example.com"]]
+    assert first_email["bcc"] == [["Bob", "bob@example.com"]]
+    assert first_email["html_body"] == "<p>ohai!</p>"
```
maymillerricci: Since this is coming from the factory, I think it feels like a "magic string" and it's not clear just looking at the test itself where it's coming from. (And like if someone were to change the factory, this test would fail.) Thoughts on doing something to make it more clear?

test/lib/bamboo/plug/sent_email_viewer_plug_test.exs
```diff
@@ -190,6 +190,44 @@ defmodule Bamboo.SentEmailViewerPlugTest do
...
+    normalize_and_push_pair(:html_email)
+    conn = conn(:post, "/api/reset.json")
+
+    conn = AppRouter.call(conn, nil)
+
+    assert conn.status == 200
+    assert {"content-type", "application/json; charset=utf-8"} in conn.resp_headers
+
+    json = Bamboo.json_library().decode!(conn.resp_body)
+    assert json == %{"ok" => true}
```
maymillerricci: Would it make sense in here to have an assertion that the emails got cleared out?

maymillerricci:
Thanks @tomtaylor! I can see arguments for both sides, but I do think having them separate would be slightly clearer, and people can decide which endpoints they want to expose. I also left you some minor comments about the tests. Thanks for your work on this!

lib/bamboo/plug/sent_email_api/sent_email_api_plug.ex
```diff
@@ -0,0 +1,112 @@
+defmodule Bamboo.SentEmailApiPlug do
+  use Plug.Router
+  alias Bamboo.SentEmail
+
+  @moduledoc """
```
maymillerricci: 👏 Yes, this documentation is great and super clear! I think it helps justify this being separate and clearly states out the case where you'd use these endpoints, vs. the ones in the `Bamboo.SentEmailViewerPlug`. Thanks for the level of detail and examples in here!

lib/bamboo/plug/sent_email_api/sent_email_api_plug.ex
```diff
@@ -0,0 +1,112 @@
...
+          # If using Plug.Router, make sure to add the `to`
+          forward "/sent_emails_api", to: Bamboo.SentEmailApiPlug
+        end
+      end
+
+  ## API endpoints
+
+  There are two endpoints:
+
+  * `GET /emails.json` exposes an array of sent emails
```
maymillerricci: Replace `array` with `list` here and the other places it's used?

lib/bamboo/plug/sent_email_api/sent_email_api_plug.ex
```diff
@@ -0,0 +1,112 @@
...
+    "html_body": "<p>hello world</p>",
+    "headers": {},
+    "from": [null, "from@example.com"],
+    "cc": [[null, "cc@example.com"]],
+    "bcc": [["Bob", "bob@example.com"]],
+    "attachments": []
+  }
+  ```
+
+  Email addresses are return in array pairs, with the first element as the
```
maymillerricci: "are return" -> "are returned"

tomtaylor:
@maymillerricci just pushed a replacement commit with the final tweaks!

tomtaylor:
Thanks!

PR #458, opened 2019-02-28
Add ability to bypass list management for SendGridAdapter

tomtaylor:
SendGrid offers the option to mark emails as being able to bypass list management (suppression of unsubscribed/bounced emails). This is useful for emails which are compulsory to send, regardless of marketing/personal preferences, like Terms of Service updates, or security notices.

This PR adds a `with_bypass_list_management` helper to enable this.


lib/bamboo/adapters/send_grid_helper.ex
```diff
@@ -135,6 +136,15 @@ defmodule Bamboo.SendGridHelper do
     raise "expected the asm_group_id parameter to be an integer, got #{asm_group_id}"
   end
 
+  def with_bypass_list_management(email, enabled) when is_boolean(enabled) do
```
maymillerricci: Thoughts on adding some small documentation with an example here to demonstrate how to use this?

maymillerricci: Thoughts on adding some small documentation with an example here to demonstrate how to use this?

test/lib/bamboo/adapters/sendgrid_helper_test.exs
```diff
@@ -138,4 +138,15 @@ defmodule Bamboo.SendGridHelperTest do
       email |> with_asm_group_id("1234")
     end
   end
+
+  test "with_bypass_list_management/2 add the correct property", %{email: email} do
```
maymillerricci: "add" -> "adds"

tomtaylor:
@maymillerricci I've fixed that typo and added some documentation. Hopefully good to go!

PR #462, opened 2019-03-07
Google Analytics settings for SendGridAdapter

tomtaylor:
Added the ability to enable/disable Google Analytics settings for an individual email in SendGrid.

### DIFF
--- a/lib/bamboo/adapters/send_grid_helper.ex
+++ b/lib/bamboo/adapters/send_grid_helper.ex
@@ -186,7 +186,6 @@ defmodule Bamboo.SendGridHelper do
       when is_boolean(enabled) do
     utm_params =
       utm_params
-      |> Enum.into(%{})
       |> Map.take(@allowed_google_analytics_utm_params)
 
     email
--- a/lib/bamboo/adapters/send_grid_helper.ex
+++ b/lib/bamboo/adapters/send_grid_helper.ex
@@ -16,6 +16,9 @@ defmodule Bamboo.SendGridHelper do
   @categories :categories
   @asm_group_id :asm_group_id
   @bypass_list_management :bypass_list_management
+  @google_analytics_enabled :google_analytics_enabled
+  @google_analytics_utm_params :google_analytics_utm_params
+  @allowed_google_analytics_utm_params ~w(utm_source utm_medium utm_campaign utm_term utm_content)a
 
   @doc """
   Specify the template for SendGrid to use for the context of the substitution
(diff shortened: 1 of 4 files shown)


lib/bamboo/adapters/send_grid_helper.ex
```diff
@@ -161,6 +164,32 @@ defmodule Bamboo.SendGridHelper do
...
+  ## Example
+
+      email
+      |> with_google_analytics(true, %{utm_source: "email", utm_campaign: "campaign"})
+
+      email
+      |> with_google_analytics(false)
+  """
+  def with_google_analytics(email, enabled, utm_params \\ %{})
+  def with_google_analytics(email, enabled, utm_params) when is_boolean(enabled) and is_map(utm_params) do
```
maymillerricci: What would you think about adding a couple of unit tests for this function that verify for example the right types being passed in and that UTM params not in white list would get filtered out?

tomtaylor:
@maymillerricci all done! I also added the ability to set the UTM params using a Keyword list, since it seemed like the flexibility would be helpful.

test/lib/bamboo/adapters/send_grid_helper_test.exs
```diff
@@ -149,4 +149,59 @@ defmodule Bamboo.SendGridHelperTest do
       email |> with_bypass_list_management(1)
     end
   end
+
+  test "with_google_analytics/2 with Map utm_params", %{email: email} do
```
maymillerricci: Should these all be `with_google_analytics/3`?

maymillerricci:
Thanks @tomtaylor! Although the flexibility for being able to pass in the params as a map or keyword list might be nice, I wonder in general whether for simplicity on this end, we might want to restrict it to only pass them as a map, and then people on their end can update what's being passed in into a map? I think in most places here similar params are expected to be maps. What do you think? Do you think it's important to have the flexibility of being able to also pass in a keyword list?

tomtaylor:
If there's precedent for just accepting Maps, that make sense. The flexibility isn't a major issue here. Will update the test descriptions when I get a moment!

tomtaylor:
@maymillerricci I think this is ready to go.

lib/bamboo/adapters/send_grid_helper.ex
```diff
@@ -161,6 +164,39 @@ defmodule Bamboo.SendGridHelper do
...
+  optionally set the UTM parameters for it. This is useful if you need to
+  control UTM tracking parameters on an individual email basis.
+
+  ## Example
+
+      email
+      |> with_google_analytics(true, %{utm_source: "email", utm_campaign: "campaign"})
+
+      email
+      |> with_google_analytics(true, utm_source: "email", utm_campaign: "campaign")
```
maymillerricci: Thanks for updating this!! Does this example now need be removed?

lib/bamboo/adapters/send_grid_helper.ex
```diff
@@ -161,6 +164,39 @@ defmodule Bamboo.SendGridHelper do
...
+      utm_params
+      |> Map.take(@allowed_google_analytics_utm_params)
+
+    email
+    |> Email.put_private(@google_analytics_enabled, enabled)
+    |> Email.put_private(@google_analytics_utm_params, utm_params)
+  end
+
+  def with_google_analytics(_email, _enabled, _utm_params) do
+    raise "expected with_google_analytics enabled parameters to be a boolean"
```
maymillerricci:
```suggestion
    raise "expected with_google_analytics enabled parameter to be a boolean"
```

tomtaylor:
Good spot! Fixed.

PR #465, opened 2019-03-26
Clarify changes and test phases in mailer test

lancejjohnson:
The variance between tests in the mailer test case all relate to the
changes in adapters and what those adapters support. Currently each
change in adapter includes a new mailer, a new config declaration, and a
new addition to the environment. As all of these changes are isolated
near the top of the file, it is less clear when reading tests lower in
the file why the response of one test is expected to differ from that of
another; there is too much separation between test setup, execution, and
assertions.

This commit reduces the amount of setup by changing only the adapter
through the use of tags and ExUnit's setup macro. This allows tests that
pivot on a particular adapter to define the adapter and declare it's use
right next to the tests that use it, bringing setup, execution, and
assertions closer together. This commit also brings common case tests to
the top of the file and groups non-common tests with describe blocks.

### DIFF
--- a/test/lib/bamboo/mailer_test.exs
+++ b/test/lib/bamboo/mailer_test.exs
@@ -2,193 +2,36 @@ defmodule Bamboo.MailerTest do
   use ExUnit.Case
   alias Bamboo.Email
 
-  defmodule FooAdapter do
-    def deliver(email, config) do
-      send(:mailer_test, {:deliver, email, config})
-    end
+  @mailer_config adapter: __MODULE__.DefaultAdapter, foo: :bar
+
+  setup context do
+    config =
+      Keyword.merge(@mailer_config, [adapter: context[:adapter]], fn
+        _, v, nil -> v
+        _, _, v -> v
+      end)
+
+    Application.put_env(:bamboo, __MODULE__.Mailer, config)
+    Process.register(self(), :mailer_test)
+    on_exit(fn -> Application.delete_env(:bamboo, __MODULE__.Mailer) end)
+    :ok
+  end
+
+  defmodule(Mailer, do: use(Bamboo.Mailer, otp_app: :bamboo))
+
+  defmodule DefaultAdapter do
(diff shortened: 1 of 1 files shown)


test/lib/bamboo/mailer_test.exs
```diff
@@ -261,6 +193,67 @@ defmodule Bamboo.MailerTest do
...
+      end
+    end
+
+    @tag adapter: AdapterWithoutAttachmentSupport
+    test "does not raise if no attachments are on the email" do
+      email = new_email(to: "foo@bar.com")
+      Mailer.deliver_now(email)
+    end
+
+    test "does not raise if adapter supports attachments" do
```
maymillerricci: Do you think it might be worth it to tag this one explicitly as using the default adapter to distinguish it? Or alternatively move this test up into the section with the tests that use the default adapter?

test/lib/bamboo/mailer_test.exs
```diff
@@ -2,250 +2,182 @@ defmodule Bamboo.MailerTest do
...
     email = new_email(from: 1)
 
     assert_raise Protocol.UndefinedError, fn ->
-      FooMailer.deliver_now(email)
+      Mailer.deliver_now(email)
     end
   end
 
-  test "raises if all recipients are nil" do
+  test "raises if all receipients are nil" do
```
maymillerricci: I think `recipients` is correct?

test/lib/bamboo/mailer_test.exs
lancejjohnson: Good suggestion. I think I'll explicitly declare the adapter but leave it in this section as we have a grouping around attachment behavior.

test/lib/bamboo/mailer_test.exs
lancejjohnson: Doh! Good catch! Not sure why I even changed that. Thanks.

PR #466, opened 2019-03-28
Add option to deliver_now to return adapter response

### LINKED ISSUE #451: Response values from adapter (deliver_now) behavior

maymillerricci:
As an alternative to #452, to resolve #451 and #161, this adds a `response: true` option to `deliver_now` that returns a tuple of the response from the adapter's `deliver` call and the `Email` struct. This is a non-breaking change, since without the `response: true` option everything will work as usual, and the `Email` will be returned, but this adds the option to pass that additional argument in and also get the response back.

### DIFF
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -64,9 +64,9 @@ defmodule Bamboo.Mailer do
   defmacro __using__(opts) do
     quote bind_quoted: [opts: opts] do
       @spec deliver_now(Bamboo.Email.t()) :: Bamboo.Email.t()
-      def deliver_now(email) do
+      def deliver_now(email, opts \\ []) do
         config = build_config()
-        Bamboo.Mailer.deliver_now(config.adapter, email, config)
+        Bamboo.Mailer.deliver_now(config.adapter, email, config, opts)
       end
 
       @spec deliver_later(Bamboo.Email.t()) :: Bamboo.Email.t()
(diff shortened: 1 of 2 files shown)


lib/bamboo/mailer.ex
```diff
@@ -113,7 +123,21 @@ defmodule Bamboo.Mailer do
...
+  def deliver_now(adapter, email, config, response: true) do
+    email = email |> validate_and_normalize(adapter)
+
+    if email.to == [] && email.cc == [] && email.bcc == [] do
+      debug_unsent(email)
+      email
+    else
+      debug_sent(email, adapter)
+      response = adapter.deliver(email, config)
+      {response, email}
```
lancejjohnson: How did you decide to order the tuple response first then email? At first glance I think I may have expected the order reversed. That's not to suggest that you _should_ reverse the order; I just thought it would be interesting to talk through.

maymillerricci: Great question! I initially decided to do `{response, email}` since I was thinking of the typical `{:ok, thing}` / `{:error, thing}` pattern and was thinking of `:ok` and `:error` as a response, so the response here would match that and come first. But now that I think about it more, it's different since it's more than just an atom with like a response status, and it comes after the sending of the email, so it probably makes more sense to come second. Plus, since this is an addition to the existing function which just returns an email, it also probably makes sense for the extra information to come after the expected information. I think I will probably reverse it. Thanks!

maymillerricci:
I reversed the tuple to be `{email, response}` in 3a98e506ae763a1edd88d61a6bc068d273ed7f53.

lancejjohnson:
@maymillerricci LGTM

PR #468, opened 2019-04-01
Cleanup and clarify documentation

lancejjohnson:
While reading through Bamboo's source code, there were a few places in
the documentation I found that could be clearer or had minor typos. This
commit aims to clarify some of the documentation and fix minor typos
here and there.

### DIFF
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -1,55 +1,56 @@
 defmodule Bamboo.Mailer do
   @moduledoc """
-  Sets up mailers that make it easy to configure and swap adapters.
+  Functions for delivering emails using adapters and delivery strategies.
 
-  Adds `deliver_now/1` and `deliver_later/1` functions to the mailer module it is used by.
+  Adds `deliver_now/1` and `deliver_later/1` functions to the mailer module it
+  is used by.
 
-  ## Bamboo ships with the following adapters
+  Bamboo [ships with several adapters][available-adapters]. It is also possible
+  to create your own adapter.
 
-  * `Bamboo.MandrillAdapter`
-  * `Bamboo.LocalAdapter`
-  * `Bamboo.TestAdapter`
-  * or create your own with `Bamboo.Adapter`
+  See the ["Getting Started" section of the README][getting-started] for an
+  example of how to set up and configure a mailer for use.
+
+  [available-adapters]: https://github.com/thoughtbot/bamboo/tree/master/lib/bamboo/adapters
+  [getting-started]: https://hexdocs.pm/bamboo/readme.html#getting-started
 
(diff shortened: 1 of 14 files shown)


lib/bamboo/mailer.ex
```diff
@@ -1,55 +1,18 @@
...
 
-  ## Bamboo ships with the following adapters
+  Bamboo [ships with several adapters][available-adapters]. It is also possible
+  to create your own adapter.
 
-  * `Bamboo.MandrillAdapter`
-  * `Bamboo.LocalAdapter`
-  * `Bamboo.TestAdapter`
-  * or create your own with `Bamboo.Adapter`
+  See the ["Getting Started" section of the README][getting-started] for an
```
maymillerricci: I wonder if it might be good to keep the complete instructions for creating a mailer here, in addition to the README, so that the Hex docs are complete by themselves? Thoughts?

lib/bamboo/plug/sent_email_api/sent_email_api_plug.ex
```diff
@@ -3,12 +3,13 @@ defmodule Bamboo.SentEmailApiPlug do
   alias Bamboo.SentEmail
 
   @moduledoc """
-  A plug that can be used in your router to expose delivered emails over a
-  JSON API. This is useful for an integration test runner that doesn't have
+  A plug that to expose delivered emails over a JSON API.
```
maymillerricci: Extra word here?

lib/bamboo/mailer.ex
jutonz: I like that idea as well. If we're concerned about maintaining two versions of documentation, we could have the readme link out to the official docs.

lib/bamboo/plug/sent_email_api/sent_email_api_plug.ex
lancejjohnson: Thanks for catching that mistake. I'll get it updated.

lib/bamboo/mailer.ex
lancejjohnson: Thanks for the feedback. I'll pull the instructions back in here and consider linking from the README to this. As you all think it's valuable to have here, my initial inclination is to duplicate the instruction as I think it's helpful for someone to see it laid out in the README as that's the first place they will likely go in trying to get started on a project. I'll play around with it and see.

lancejjohnson: Having looked at, I think some duplication is the best solution. I've re-used the code examples from the README while updating the text flow to fit better within this context.

jutonz: Sweet, works for me! Thanks for doing some investigation 🎉

PR #475, opened 2019-04-18
Adding content_id for displaying attachements in HTML email

ttyerl:
Adding content_id for displaying attachements in HTML email. The corresponding changes to bamboo_smtp have been made and a PR created for it.

### DIFF
--- a/lib/bamboo/attachment.ex
+++ b/lib/bamboo/attachment.ex
@@ -8,7 +8,8 @@ defmodule Bamboo.Attachment do
           path: nil | String.t(),
           filename: nil | String.t(),
           content_type: nil | String.t(),
-          data: nil | binary()
+          data: nil | binary(),
+          content_id: nil | String.t()
         }
 
   @doc ~S"""
@@ -44,7 +45,14 @@ defmodule Bamboo.Attachment do
     content_type = opts[:content_type] || determine_content_type(path)
     content_id = opts[:content_id]
     data = File.read!(path)
-    %__MODULE__{path: path, data: data, filename: filename, content_type: content_type, content_id: content_id}
+
+    %__MODULE__{
+      path: path,
+      data: data,
+      filename: filename,
+      content_type: content_type,
+      content_id: content_id
+    }
   end
 
   defp determine_content_type(path) do
--- a/lib/bamboo/attachment.ex
+++ b/lib/bamboo/attachment.ex
@@ -2,7 +2,7 @@ defmodule Bamboo.Attachment do
   @moduledoc """
   """
 
-  defstruct filename: nil, content_type: nil, path: nil, data: nil
+  defstruct filename: nil, content_type: nil, path: nil, data: nil, content_id: nil
 
   @type t :: %__MODULE__{
           path: nil | String.t(),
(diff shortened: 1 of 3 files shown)


ttyerl:
It says "mix format --check-formatted" failed but gives me no indication of why it fails.

maymillerricci:
@ttyerl Thanks for this! 

If you run `mix format` locally, it should fix whatever formatting issues there are.

Also, is there a test we can add for this new behavior?

ttyerl:
I ran this locally and everything passed. Not too sure how to fix this failing test. Hints appreciated.

lib/bamboo/attachment.ex
```diff
@@ -2,24 +2,36 @@ defmodule Bamboo.Attachment do
...
           content_type: nil | String.t(),
-          data: nil | binary()
+          data: nil | binary(),
+          content_id: nil | String.t()
         }
 
   @doc ~S"""
   Creates a new Attachment
 
+  context_id can be use to embed an image, attach it and reference in the message body by
```
maymillerricci: `context_id` -> `content_id`
`can be use` -> `can be used`
`and reference` -> `and reference it`

lib/bamboo/attachment.ex
```diff
@@ -2,24 +2,36 @@ defmodule Bamboo.Attachment do
...
+  setting its CID (Content-ID) and using a standard HTML tag:
+
+    <img src="cid:some-image-cid" alt="img" />
+
+  within a HTML email message.
```
maymillerricci: `within a` -> `within an`

lib/bamboo/attachment.ex
```diff
@@ -2,24 +2,36 @@ defmodule Bamboo.Attachment do
...
+          content_id: nil | String.t()
         }
 
   @doc ~S"""
   Creates a new Attachment
 
+  context_id can be use to embed an image, attach it and reference in the message body by
+  setting its CID (Content-ID) and using a standard HTML tag:
+
+    <img src="cid:some-image-cid" alt="img" />
```
maymillerricci: I think this needs 2 more spaces of indentation in order to render correctly as code on the docs.

lib/bamboo/attachment.ex
```diff
@@ -2,24 +2,36 @@ defmodule Bamboo.Attachment do
...
 
       Bamboo.Attachment.new("/path/to/attachment.png")
       Bamboo.Attachment.new("/path/to/attachment.png", filename: "image.png")
-      Bamboo.Attachment.new("/path/to/attachment.png", filename: "image.png", content_type: "image/png")
+      Bamboo.Attachment.new("/path/to/attachment.png", filename: "image.png", content_type: "image/png", content_id: "12387432")
       Bamboo.Attachment.new(params["file"]) # Where params["file"] is a %Plug.Upload
+
+      email
+        |> put_html_layout({LayoutView, "email.html"})
+        |> put_attachment(%Bamboo.Attachment{content_type: "image/png", filename: "logo.png", data: get_content(), content_id: "2343333333"})
```
maymillerricci: The code formatter would line the pipes up directly with `email` - so all the lines are directly in line. Also, what do you think about switching the `data` passed in in the example to some arbitrary data instead of referencing a `get_content()` function, to be as understandable as possible?

maymillerricci:
I made these changes and merged in https://github.com/thoughtbot/bamboo/commit/228827b331673623eb828ae3225818502c681c82.

PR #476, opened 2019-04-18
Support Mailgun EU region

KristerV:
Mailgun now has two regions: US (default) and EU. Much due to GDPR regulations. https://www.mailgun.com/regions

I'm open for further suggestions.

### DIFF
--- a/lib/bamboo/adapters/mailgun_adapter.ex
+++ b/lib/bamboo/adapters/mailgun_adapter.ex
@@ -11,7 +11,8 @@ defmodule Bamboo.MailgunAdapter do
       config :my_app, MyApp.Mailer,
         adapter: Bamboo.MailgunAdapter,
         api_key: "my_api_key" # or {:system, "MAILGUN_API_KEY"},
-        domain: "your.domain" # or {:system, "MAILGUN_DOMAIN"}
+        domain: "your.domain" # or {:system, "MAILGUN_DOMAIN"},
+        region: "eu" # or {:system, "MAILGUN_DOMAIN"}, (leave blank for US)
 
       # Define a Mailer. Maybe in lib/my_app/mailer.ex
       defmodule MyApp.Mailer do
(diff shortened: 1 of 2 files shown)


lib/bamboo/adapters/mailgun_adapter.ex
```diff
@@ -20,7 +21,6 @@ defmodule Bamboo.MailgunAdapter do
   """
 
   @service_name "Mailgun"
-  @base_uri "https://api.mailgun.net/v3"
```
maymillerricci: What do you think about keeping this as a constant here and then referencing the constant below?

lib/bamboo/adapters/mailgun_adapter.ex
```diff
@@ -81,7 +81,14 @@ defmodule Bamboo.MailgunAdapter do
   def supports_attachments?, do: true
 
   defp full_uri(config) do
-    Application.get_env(:bamboo, :mailgun_base_uri, @base_uri) <>
+    uri =
+      if Map.has_key?(config, :region) do
+        "https://api.#{config.region}.mailgun.net/v3"
+      else
+        "https://api.mailgun.net/v3"
+      end
```
maymillerricci: Thoughts on extracting this conditional into its own function?

arjan:
Oops, I totally overlooked this PR. #491 is an alternative approach to this.

KristerV:
ah.. I've got this in my (short) todo list but still haven't gotten around to it. Your change is more complicated, yet more flexible against future change too. Plus you've tested it and written proper documentation. Not even sure which one i'd prefer, let the admins decide i guess :)

maymillerricci:
Thanks @KristerV and @arjan! Yeah, at this point I think I'm leaning towards #491 which has a little bit more functionality and has good documentation and test coverage. Thanks for your work on this!

KristerV:
Okay, closing this.

Issue #445, opened 2018-12-10
SendGrid headers


bjunc:
It doesn't appear that the SendGrid adapter supports headers.  Is that correct?  I see one-off support for a `reply-to`, but nothing anything else (eg. `X-Foo`).

Just to test my suspicion, I copied/modified the SendGridAdapter to merge `%Email{headers: headers}` into the body; which worked.  Happy to submit a PR, if headers are in-fact unsupported at the moment.

maymillerricci:
@bjunc Thanks for posting! Which header(s) are you looking to pass through? Or were you thinking of making it possible to pass anything through, that map to the `headers` parameter for the API call? I think `reply-to` has one-off handling since it maps to a specific `reply_to` param. In any case, please do open a PR if there's something you are wanting for this that we don't yet have!

michaelm244:
@maymillerricci I have this same problem - the headers that I'm interested in passing through are `In-Reply-To` and `References`. This allows you to reply to an email directly, assuming you have the message-ID of the email you'd like to reply to.

I think it's better to allow the user of the library to define what headers they want to send, rather than build one-off support for each specific header.

I just opened up a PR that adds this functionality, would love to hear any feedback/thoughts: https://github.com/thoughtbot/bamboo/pull/479

bjunc:
Our use-case is similar. We're adding `In-Reply-To` as well as `References` to allow for message threading. We implemented something similar to the PR. Although, we simply put `put_headers` first in the pipeline so we don't worry about the `put_reply_to` (letting the user decide if they want to override the headers previously set).  For example:

```elixir
defp to_sendgrid_body(%Email{} = email, config) do
    %{}
    |> put_headers(email)
    |> put_from(email)
    |> put_personalization(email)
    |> put_reply_to(email)
    |> put_subject(email)
    |> put_content(email)
    |> put_template_id(email)
    |> put_attachments(email)
    |> put_categories(email)
    |> put_settings(config)
  end

defp put_headers(body, %Email{headers: headers})  when is_map(headers), 
  do: Map.put(body, :headers, headers)
```

Since the SendGrid docs go out of their way to say not to send certain reserved headers, we might want to make sure those aren't included in the header map (per the PR comment).

maymillerricci:
This was implemented in #479.

PR #479, opened 2019-05-06
Send headers in Sendgrid Adapter

michaelm244:
Implements #445. Allows arbitrary setting of Sendgrid email headers through `Bamboo.Email`'s `headers` field.

Detailed documentation on Sendgrid's support for email headers here: https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/index.html

### DIFF
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -10,6 +10,9 @@ defmodule Bamboo.SendGridAdapter do
 
       put_header("reply-to", "foo@bar.com")
 
+  To set arbitrary email headers, set them in the `headers` property of the [Bamboo.Email](Bamboo.Email) struct.
+  Note that some header names are reserved for use by Sendgrid.
+
   ## Example config
 
       # In config/config.exs, or config.prod.exs, etc.
@@ -98,6 +101,7 @@ defmodule Bamboo.SendGridAdapter do
     |> put_from(email)
     |> put_personalization(email)
     |> put_reply_to(email)
+    |> put_headers(email)
     |> put_subject(email)
     |> put_content(email)
     |> put_template_id(email)
@@ -160,6 +164,14 @@ defmodule Bamboo.SendGridAdapter do
   end
 
   defp put_reply_to(body, _), do: body
+
+  defp put_headers(body, %Email{headers: headers}) when is_map(headers) do
+    prepared_headers = Map.delete(headers, "reply-to")
+
+    Map.put(body, :headers, prepared_headers)
+  end
+
+  defp put_headers(body, _), do: body
 
   defp put_subject(body, %Email{subject: subject}) when not is_nil(subject),
     do: Map.put(body, :subject, subject)
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -168,7 +168,8 @@ defmodule Bamboo.SendGridAdapter do
 
   defp put_headers(body, %Email{headers: headers}) when is_map(headers) do
     headers_without_tuple_values =
-      Map.delete(headers, "reply-to")
+      headers
+      |> Map.delete("reply-to")
       |> Map.delete("Reply-To")
 
     Map.put(body, :headers, headers_without_tuple_values)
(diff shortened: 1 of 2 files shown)


lib/bamboo/adapters/send_grid_adapter.ex
```diff
@@ -148,6 +152,14 @@ defmodule Bamboo.SendGridAdapter do
 
   defp put_reply_to(body, _), do: body
 
+  defp put_headers(body, %Email{headers: headers}) when is_map(headers) do
+    prepared_headers = Map.delete(headers, "reply-to")
```
michaelm244: From the [Sendgrid docs](https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/index.html):
```
Must not be any of the following reserved headers: x-sg-id, x-sg-eid, received, dkim-signature, Content-Type, Content-Transfer-Encoding, To, From, Subject, Reply-To, CC, BCC
```

Would it be worth it to explicitly delete these reserved keys (e.g. `Subject`) here? Or log a warning if they're passed? Or just link to the Sendgrid documentation on reserved headers in the `moduledoc`?

michaelm244:
@maymillerricci could I get a review on this?

lib/bamboo/adapters/send_grid_adapter.ex
maymillerricci: @michaelm244 Thanks for the PR! This is a great question. Do we know what happens if you try to send an email with one of these reserved headers? Does it just get ignored? I like the idea of logging something to notify the user, and definitely linking to the docs. If it just gets ignored it may not be necessary to delete it, but if there's some other effect or if it causes an error then we would want to delete it. What are your thoughts?

michaelm244: Just tested out by adding a header with the key of `Subject` - Sendgrid returns a 500 response with a very informative error message:
```
{
  "errors": [
    {
      "message": "Header can not be one of the reserved keys. Refer to documentation link.",
      "field": "headers.Subject",
      "help": "http://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/errors.html#message.headers"
    }
  ]
}
```

Since the Sendgrid API enforces it well, is it necessary for this library to enforce it also? The user will have very clear feedback of what's going wrong if they try to send a disallowed header. If the library enforces it, it will also take on the burden of ensuring that the library's list of disallowed headers is always up-to-date with Sendgrid's.

michaelm244: @maymillerricci also just linked to Sendgrid docs in the `@moduledoc`

michaelm244:
@maymillerricci any updates on this? anything I should change?

lib/bamboo/adapters/send_grid_adapter.ex
maymillerricci: @michaelm244 So sorry for the delay! Yes, I agree that since the user will get clear feedback then we don't need to enforce it on this end, and we don't want the extra burden of keeping the lists lined up. This looks great! My only other thought is: do you want to add a test for this special case of the `reply-to` header that's being deleted since it is handled elsewhere?

stratigos:
@michaelm244 and @maymillerricci I attempted to add the suggested test, and ended up having to do the following:

* d584ca1 rebases `thoughtbot/bamboo:master` into this branch
* 079ce95 fixes some code formatting issues I created fixing merge conflicts
* b825220 fixes a test that broke after rebasing with upstream
  * a test was introduced which added a `"Reply-To"` header with a Tuple value, which breaks `Jason.Encoder`
  * i noticed that `put_headers/2` was already handling something with the `"reply-to"` header, and guessed that I'd need to do the same for `"Reply-To"`
  * i then renamed the variable from `prepared_headers` to `headers_without_tuple_values` to reflect that removing Tuple values is necessary for adding headers
* 541af34 finally adds the requested test for removing `"reply-to"` and `"Reply-To"` headers

I think this may be ready to merge now, what do you think?

lib/bamboo/adapters/send_grid_adapter.ex
```diff
@@ -161,6 +166,16 @@ defmodule Bamboo.SendGridAdapter do
 
   defp put_reply_to(body, _), do: body
 
+  defp put_headers(body, %Email{headers: headers}) when is_map(headers) do
+    headers_without_tuple_values =
+      Map.delete(headers, "reply-to")
+      |> Map.delete("Reply-To")
```
maymillerricci: Tiny code style thing. My parallel structure brain would find it easier to understand at a glance if it were written like: 
```
headers
|> Map.delete("reply-to")
|> Map.delete("Reply-To")
```

What do you think?

stratigos: Thanks @maymillerricci, I think that looks better, is more consistent with the code style, and feels more idiomatic. :alembic: :+1:

michaelm244:
Thanks for taking this to the finish line @stratigos 🙂

Issue #444, opened 2018-12-03
assert_email_delivered_with unintuitive test syntax


blacksph3re:
To test on certain fields it requires a weird pattern matching, as the passed keywords are matched directly against the mail structure. This makes regex matching for these fields impossible, as the field of the mail structure is passed directly to regex.match

Concretely it is:

- from: `from: {nil: "from@address.com"}`
- html_body: `html_body: {:ok, "<some>html</some>"}`
- to: `to: [nil: "to@address.com"]`

I would find it more intuitive if I wouldn't have to insert those maps and lists, also regex matching would work then. If I try with the intuitive way (`assert_email_delivered_with(to: "to@address.com")`) I get the error:

```
The parameters given do not match.
     
       Parameters given:
     
         [to: "to@address.com"]
     
       Email recieved:
     
         %{assigns: :assigns_removed_for_testing, attachments: [], bcc: [], cc: [], from: {nil, "from@address.com"}, headers: %{}, html_body: {:ok, "<!-- some html-->"}, private: %{}, subject: "some subject", text_body: nil, to: [nil: "to@address.com"]}

```

lancejjohnson:
@blacksph3re Thanks for submitting this issue recommending a possible improvement. Would you be willing to provide an example test that uses the syntax as it stands followed by the same test with the syntax that you're recommending? I think that will help clarify the request.

tcoopman:
Some examples:

```elixir
# instead of
assert_email_delivered_with(from: {nil, "foo@bar.com"})
assert_email_delivered_with(to: [nil, "foo@bar.com"])
# be able to use
assert_email_delivered_with(from: "foo@bar.com")
assert_email_delivered_with(to: "foo@bar.com")
```

germsvel:
It seems like this was resolved by https://github.com/thoughtbot/bamboo/pull/482, so I'm going to close this. 

If this is still an issue, please let me know. Happy to reopen it.

randycoulman:
I realize this issue has been closed for a while, but I just upgraded to v2.2.0 and checked to see if this was fixed for us. It turns out it isn't fixed by #482.

I investigated, and it looks like `Bamboo.Formatter` isn't having the desired effect:

```elixir
my_app (1)> Bamboo.Formatter.format_email_address([nil: "someone@example.com"], %{type: :to})
[nil: "someone@example.com"]
my_app (2)> Bamboo.Formatter.format_email_address({nil, "someone@example.com"}, %{type: :from})
[...]

PR #482, opened 2019-05-27
Use a formatter for `assert_email_delivered_with/1`

drapergeek:
This should make it a bit easier to test emails without needing to use the specific format that bamboo uses.

### DIFF
--- a/lib/bamboo/test.ex
+++ b/lib/bamboo/test.ex
@@ -182,6 +182,12 @@ defmodule Bamboo.Test do
       unsent_email = Bamboo.Email.new_email(subject: "something else")
       assert_email_delivered_with(subject: "something else") # Will fail
 
+  The function will use the Bamboo Formatter when checking email addresse.s
+
+      email = Bamboo.Email.new_email(to: "someone@example.com")
+      email |> MyApp.Mailer.deliver
+      assert_email_delivered_with(to: "someone@example.com") # Will pass
+
   You can also pass a regex to match portions of an email.
 
   ## Example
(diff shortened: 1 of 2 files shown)


test/lib/bamboo/adapters/test_adapter_test.exs
```diff
@@ -54,6 +54,7 @@ defmodule Bamboo.TestAdapterTest do
     end
 
     sent_email |> TestMailer.deliver_now()
+    assert_email_delivered_with(from: "foo@bar.com")
```
maymillerricci: It took me a second to understand this test, and I was wondering about the idea of having it be its own (named) test, but this does follow the existing pattern of including all of these within the same test about the helpers, so maybe making changes to it would be a separate issue? Thoughts?

drapergeek: I'm not a huge fan of this style, but I wasn't going to fight against the current standard. I don't have a strong opinion in favor of changing it but I would say it's a separate issue.

drapergeek:
Hi @maymillerricci 👋!

Thanks for the review, will happily add some documentation! Yes, this will resolve #444

lib/bamboo/test.ex
```diff
@@ -182,6 +182,12 @@ defmodule Bamboo.Test do
       unsent_email = Bamboo.Email.new_email(subject: "something else")
       assert_email_delivered_with(subject: "something else") # Will fail
 
+  The function will use the Bamboo Formatter when checking email addresse.s
```
maymillerricci:
```suggestion
  The function will use the Bamboo Formatter when checking email addresses.
```

maymillerricci:
Merged in 73592c57a12b700d6c67e699b8c4645059935ff7. Thanks for this!

Issue #474, opened 2019-04-16
refute_delivered_email_with?


yuvalidan:
Something like `refute_delivered_email_with` or the ability to pass options to `refute_delivered_email` would be super helpful. Currently I have a problem when testing emails where the html includes unique elements (like a timestamp or a token) because it would never match the email that's generated in the test.

I can make the affirmative tests work with `assert_email_delivered_with`, but have no way of testing that a similar email was *not* sent because currently only full matches are supported.

maymillerricci:
This was implemented in #484.

PR #484, opened 2019-06-07
Add refute_email_delivered_with helper, tests

mtarnovan:
See https://github.com/thoughtbot/bamboo/issues/474

Let me know if you want to see more tests or if i missed anything obvious. Thanks.

### DIFF
--- a/lib/bamboo/test.ex
+++ b/lib/bamboo/test.ex
@@ -203,6 +203,26 @@ defmodule Bamboo.Test do
     end
   end
 
+  defmacro refute_email_delivered_with(email_params) do
+    quote bind_quoted: [email_params: email_params] do
+      import ExUnit.Assertions
+
+      received_email_params =
+        receive do
+          {:delivered_email, email} -> Map.from_struct(email)
+        after
+          100 -> []
+        end
+
+      if is_nil(received_email_params) do
+        refute false
+      else
+        refute Enum.any?(email_params, fn {k, v} -> do_match(received_email_params[k], v) end),
+              Bamboo.Test.flunk_attributes_match(email_params, received_email_params)
+      end
+    end
+  end
+
   @doc false
   def do_match(value1, value2 = %Regex{}) do
     Regex.match?(value2, value1)
@@ -252,6 +272,21 @@ defmodule Bamboo.Test do
   def flunk_attributes_do_not_match(params_given, params_received) do
     """
     The parameters given do not match.
+
+      Parameters given:
+
+        #{inspect(params_given)}
+
+      Email received:
+
+        #{inspect(params_received)}
+    """
+  end
+
+  @doc false
+  def flunk_attributes_match(params_given, params_received) do
+    """
+    The parameters given match.
 
       Parameters given:
 
--- a/lib/bamboo/test.ex
+++ b/lib/bamboo/test.ex
@@ -203,6 +203,21 @@ defmodule Bamboo.Test do
     end
   end
 
+  @doc """
+  Check that no email was sent with the given parameters
+
+  Similarly to `assert_email_delivered_with`, the assertion waits up to 100ms before
+  failing. Note that you need to send the email again if you want to make other
+  assertions after this, as this will receive the `{:delivered_email, email}` message.
+
+  ## Examples
+
+      Bamboo.Email.new_email(subject: "something") |> MyApp.Mailer.deliver
+      refute_email_delivered_with(subject: "something else") # Will pass
+
+      email = Bamboo.Email.new_email(subject: "something") |> MyApp.Mailer.deliver
+      refute_email_delivered_with(subject: ~r/some/") # Will fail
+  """
   defmacro refute_email_delivered_with(email_params) do
     quote bind_quoted: [email_params: email_params] do
       import ExUnit.Assertions
--- a/lib/bamboo/test.ex
+++ b/lib/bamboo/test.ex
@@ -218,7 +218,7 @@ defmodule Bamboo.Test do
         refute false
       else
         refute Enum.any?(email_params, fn {k, v} -> do_match(received_email_params[k], v) end),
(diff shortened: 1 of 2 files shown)


lib/bamboo/test.ex
```diff
@@ -203,6 +203,26 @@ defmodule Bamboo.Test do
     end
   end
 
+  defmacro refute_email_delivered_with(email_params) do
```
maymillerricci: Thanks for this!! Thoughts on adding some documentation for it?

test/lib/bamboo/adapters/test_adapter_test.exs
```diff
@@ -172,6 +172,38 @@ defmodule Bamboo.TestAdapterTest do
...
+  test "refute_email_delivered_with when email does not match" do
+    new_email(
+      to: {nil, "foo@bar.com"},
+      from: {nil, "foo@bar.com"},
+      text_body: "I really like coffee"
+    )
+    |> TestMailer.deliver_now()
+
+    refute_email_delivered_with(text_body: ~r/tea/)
+    refute_email_delivered_with(to: "bla")
```
maymillerricci: I think there can only be one assertion of this type per email delivery -- the received email params are only pulled for the first assertion. If you switch this line to `refute_email_delivered_with(to: [nil: "foo@bar.com"])` it passes which it should not? So I think for these additional assertions to be relevant, we need to call `deliver_now` again before each - similar to https://github.com/thoughtbot/bamboo/blob/3f79b73a4e27d9b27694fb4a72473caab46c857f/test/lib/bamboo/adapters/test_adapter_test.exs#L56-L63

mtarnovan: You're right, of course, will fix. 

I wonder though if  this is the best way to implement these assertions. It would be better to store the emails in an agent and match over all of them.  That would require a big refactor, but it might be worth it.

lib/bamboo/test.ex
mtarnovan: 249ce54

test/lib/bamboo/adapters/test_adapter_test.exs
mtarnovan: 1deb19c

lib/bamboo/test.ex
```diff
@@ -203,6 +203,41 @@ defmodule Bamboo.Test do
...
+  failing. Note that you need to send the email again if you want to make other
+  assertions after this, as this will receive the `{:delivered_email, email}` message.
+
+  ## Examples
+
+      Bamboo.Email.new_email(subject: "something") |> MyApp.Mailer.deliver
+      refute_email_delivered_with(subject: "something else") # Will pass
+
+      email = Bamboo.Email.new_email(subject: "something") |> MyApp.Mailer.deliver
+      refute_email_delivered_with(subject: ~r/some/") # Will fail
```
maymillerricci:
```suggestion
      refute_email_delivered_with(subject: ~r/some/) # Will fail
```

test/lib/bamboo/adapters/test_adapter_test.exs
```diff
@@ -172,6 +172,69 @@ defmodule Bamboo.TestAdapterTest do
...
+        subject: subject = "coffee"
+      )
+
+    TestMailer.deliver_now(mail)
+    refute_email_delivered_with(subject: ~r/tea/)
+
+    TestMailer.deliver_now(mail)
+
+    try do
+      refute_email_delivered_with(subject: subject)
```
maymillerricci: This, and the others below, are in the test called `when email does not match` but it seems they are testing `when email matches`, which there is a separate test for? Thoughts on regrouping them, or alternatively removing some of them if they don't each add much additional value?

Also, if we do keep all of them, thoughts on extracting a helper function for the repeated code?

test/lib/bamboo/adapters/test_adapter_test.exs
```diff
@@ -172,6 +172,69 @@ defmodule Bamboo.TestAdapterTest do
     end
   end
 
+  test "refute_email_delivered_with when email does not match" do
+    mail =
+      new_email(
+        to: to = [{nil, "foo@bar.com"}],
```
maymillerricci: Not a big deal, but our stylistic preference for this, and `subject` below, is to assign the variable on its own line and then use it, vs. doing it inline like this, since we think it makes it easier to understand at a glance. Your call whether you want to change this!

test/lib/bamboo/adapters/test_adapter_test.exs
maymillerricci: > I wonder though if this is the best way to implement these assertions. It would be better to store the emails in an agent and match over all of them. That would require a big refactor, but it might be worth it.

What do you think about keeping it as is for now and merging this and opening a new issue for the potential refactor?

test/lib/bamboo/adapters/test_adapter_test.exs
mtarnovan: Sure 3928f3e

test/lib/bamboo/adapters/test_adapter_test.exs
mtarnovan: Right, I don't think they add too much value. I've refactored the tests a bit if you want to take another look.

maymillerricci: Nice, thanks so much for this!

lib/bamboo/test.ex
```diff
@@ -203,6 +203,41 @@ defmodule Bamboo.Test do
...
+        receive do
+          {:delivered_email, email} -> Map.from_struct(email)
+        after
+          100 -> []
+        end
+
+      if is_nil(received_email_params) do
+        refute false
+      else
+        refute Enum.any?(email_params, fn {k, v} -> do_match(received_email_params[k], v) end),
```
mtarnovan, line 235: maybe this should be `all?` instead 🤔 @mtarnovan

PR #486, opened 2019-06-10
Upgrade ex_doc

maartenvanvliet:
ExDoc has released version 0.20 that adds a lot of features such as autosave and version support

See: https://github.com/elixir-lang/ex_doc/blob/master/CHANGELOG.md#v0200


mix.exs
```diff
@@ -58,9 +58,8 @@ defmodule Bamboo.Mixfile do
       {:phoenix, "~> 1.1", only: :test},
       {:phoenix_html, "~> 2.2", only: :test},
       {:excoveralls, "~> 0.4", only: :test},
-      {:floki, "~> 0.8", only: :test},
-      {:ex_doc, "~> 0.19", only: :dev},
-      {:earmark, ">= 0.0.0", only: :dev},
+      {:floki, "~> 0.8", osnly: :test},
```
maymillerricci:
```suggestion
      {:floki, "~> 0.8", only: :test},
```

PR #487, opened 2019-06-18
SendGrid Adapter Reply-To parameter updates

cpjolicoeur:
This PR adds support for the "native" `Reply-To` header string (in addition to the already supported custom `reply-to` header string).

Additionally, it also adds full support for SendGrids `reply_to` parameter format with both the required `email` field and optional `name` field.

I split these two additions into 2 separate commits for easier review and editing.  If there is no desire to add full support for the `name` field, I can gladly drop that commit and just send across the `Reply-To` commit. 

To me, at a minimum, the `Reply-To` format of the header string should be allowed as this is the standard name for the header in SMTP spec and is referenced directly in the top level `Bamboo.Email` docs on the `put_header/3` function.

### DIFF
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -146,6 +146,10 @@ defmodule Bamboo.SendGridAdapter do
     Map.put(body, :reply_to, %{email: reply_to})
   end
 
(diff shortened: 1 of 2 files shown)


cpjolicoeur:
Any thoughts on this?

cpjolicoeur:
Is this project still maintained by Thoughbot?  Should we (users) start maintaining a separate fork?  

Let us know how we can contribute as things seem to be pretty stagnant here?  Just looking to help.

Issue #299, opened 2017-06-12
BambooMandrillAdapterApiError - Process #PID<> terminating: :timeout


casio:
Hey guys,

we're experiencing seemingly random errors in our production app with this stacktrace:
``` 
(bamboo) lib/bamboo/adapters/mandrill_adapter.ex:72: Bamboo.MandrillAdapter.deliver/2
(elixir) lib/task/supervised.ex:85: Task.Supervised.do_apply/2
(stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
``` 

We're on `0.8.0` using the Mandrill adapter. 
Have you maybe seen this elsewhere?
Or maybe, would an attempt to increase the `hackney` timeout options be a feasible attempt to workaround?

Thanks for a pointer!

casio:
fwiw I tried to fix this by configuring `hackney` likes so:
`config :hackney, timeout: 150000, max_connections: 500`

didnt help - still getting the timeouts

however, we're using an umbrella app and I added the config in the app where `hackney` is explicitly started but which is not the app sending the mail. I'll also put it in the other app & report back here

paulcsmith:
@casio any luck? You may want to try the latest beta or use master branch. If I remember correctly we updated hackney. Maybe that'd fix it

Preen:
This is a killer for us :(

paulcsmith:
I haven't seen this so I'm unsure how to fix. If someone finds a fix, please submit a PR for it. @casio Did that fix work for you?

kcurtin:
FYI pretty sure this is an issue in more recent versions of hackney where connections aren't being release.. I've ran into this problem in other elixir packages that depend on hackney.

I'm linking to a PR that attempted to address this but updating hackney did not full resolve the issues in my projects: https://github.com/benoitc/hackney/issues/420

egeersoz:
I run into this frequently when dealing with large attachments. I have a function that sends data to another service and receives a large (15 MB+) PDF file, which I then need to attach to an email and send via Bamboo. I get the PDF file without issues, but Bamboo times out while sending.

churcho:
I am also getting this `(Bamboo.ApiError) :timeout`
Can't find a way around this. I am sending out large attachments.

tmarkanov:
I took the liberty of forking the repo to resolve this issue, as I have had the same problem.
https://github.com/tmarkanov/bamboo

maymillerricci:
[...]

PR #489, opened 2019-06-28
Add hackney options support to config

tmarkanov:
Add hackney options support to config, e.g.:
```
config :agora, MyApp.Mailer,
  adapter: Bamboo.MandrillAdapter,
  api_key: "x",
  hackney: [
    recv_timeout: 5 * 60 * 1000
  ]
```
Added in order to circumvent https://github.com/thoughtbot/bamboo/issues/299, i.e. to enable setting of `recv_timeout` in order to be able to send larger attachments without errors.


maymillerricci:
Thanks for this! What do you think about adding some tests to verify that the configuration works as expected?

tmarkanov:
Sure thing! I am a bit busy at the moment but will try to do it as soon as I can.

mtarnovan:
Ups, I overlooked this PR when i opened https://github.com/thoughtbot/bamboo/pull/498

mtarnovan:
@maymillerricci @tmarkanov if that's ok with you we can close this in favor of #498 which is basically the same thing, but has more some docs. Re: May's comment about verifying the configuration works as expected, on the other PR that's done through a helper that's also tested. How would you suggest to verify the configuration actually works? Mocking hackney seems a bit overkill for this.

tmarkanov:
@mtarnovan it's fine by me. I have been very busy lately so can't really tell when I would catch some time to add some testing. Closing.

PR #496, opened 2019-08-23
Fix variable substitution for nested maps in Mailgun templates

milmazz:
The previous implementation for `Bamboo.MailgunHelper.substitute_variables/2`
seems that does not handle correctly nested maps:

```elixir
private: %{
  mailgun_custom_vars: %{
    campaign: %{name: "ex"},
    from: %{first_name: "Zoe"},
    to: %{first_name: "Theron"}
  },
  "o:tag": "notices",
  template: "thanks"
},
```

Reviewing the logs from Mailgun I saw the following JSON structure:

```json
"user-variables": {
  "from[first_name]": "Zoe",
  "to[first_name]": "Theron",
  "campaign[name]": "ex"
}
```

Which wasn't expected by the Handlebars template, but, I found that you could
use the header `X-Mailgun-Variables` to inject deeply nested maps from Bamboo, and
that's what this proposal includes.

Also, according to the Mailgun documentation when you use a template, you can also use the option: `t:text = yes` to pass the rendered template as the `text` part of the message. Should we include this option on the `MailgunHelper` module?

> | Parameter | Description |
> | ---------- | ------------ | 
> | t:text | Pass yes if you want to have rendered template in the text part of the message in case of template sending |

Regarding the last point, seems that doing `Email.put_private(:"t:text", "yes")` does not work as expected, anyone has tested this feature from Mailgun using Bamboo? I also tried using the header: `X-Mailgun-Template-Text` with `true` as value. @mstratman Have you worked with this option from Mailgun?

### DIFF
--- a/lib/bamboo/adapters/mailgun_helper.ex
+++ b/lib/bamboo/adapters/mailgun_helper.ex
@@ -6,6 +6,8 @@ defmodule Bamboo.MailgunHelper do
 
   alias Bamboo.Email
 
+  @mailgun_header_for_custom_vars "X-Mailgun-Variables"
+
   @doc """
   Add a tag to outgoing email to help categorize traffic based on some
   criteria, perhaps separate signup emails from password recovery emails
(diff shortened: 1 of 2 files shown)


lib/bamboo/adapters/mailgun_helper.ex
```diff
@@ -63,8 +73,19 @@ defmodule Bamboo.MailgunHelper do
       |> MailgunHelper.substitute_variables(%{ "greeting" => "Hello!", "password_reset_link" => "https://example.com/123" })
     
   """
-  def substitute_variables(email, variables = %{}) do
-    custom_vars = Map.get(email.private, :mailgun_custom_vars, %{})
-    Email.put_private(email, :mailgun_custom_vars, Map.merge(custom_vars, variables))
+  def substitute_variables(%Email{headers: headers} = email, variables = %{}) do
+    header_key = "X-Mailgun-Variables"
```
stratigos: How do you feel about promoting `"X-Mailgun-Variables"` to a module attribute? Perhaps something that expresses: `@mailgun_header_for_custom_vars` ?

milmazz: @stratigos Agreed. I added the module attribute as you suggested.

maymillerricci:
@milmazz Thanks so much for this! We actually just merged #504 which implements the `template_text` functionality (along with some other template-related functionality), but the `X-Mailgun-Variables` header replacement here for nested maps looks great! Can you rebase, and update this PR to just be for the `X-Mailgun-Variables` header replacement? Sorry for not getting to this sooner!

milmazz:
@maymillerricci Done.

milmazz:
@maymillerricci any updates about this?

Issue #470, opened 2019-04-02
Allow dynamic configuration in Mailer functions without directly calling through to `Bamboo.Mailer`


lancejjohnson:
See #309, #414, #282 for examples and use cases.

The current recommendation via #309 is to use the `Bamboo.Mailer` undocumented functions. As Bamboo discourages the use of direct calls to `Bamboo.Mailer`, it should provide the ability to accomplish this same task through the usable functions.

sandeshsoni:
@lancejjohnson If we should not directly call functions of `Bamboo.Mailer`, should an another module be created that makes the call? 

I am checking if I can help and will be happy to contribute!

lancejjohnson:
@sandeshsoni What I'm referring to above and probably should have clarified is that `deliver_now/1` and `deliver_later/1` on `Bamboo.Mailer` raise an exception. This is designed, I believe, to force you to define your own mailer (e.g. `MyApp.Mailer`) and use the macro `use Bamboo.Mailer` within it.  The recommendation above is to allow the work around (i.e. calling the undocumented 3 arity functions) behavior to be available through the functions provided by `use Bamboo.Mailer`. I hope that helps clarify.

stratigos:
I believe this is now satisfied by #497 on the latest `master`

PR #497, opened 2019-09-12
Allow Dynamic Override of Mailer Config

stratigos:
:woman_technologist: Bamboo's adopters would like to be able to pass dynamic configuration
into a single email delivery call, for scenarios such as un/pw, port
changes, or multi-tenancy.

:mailbox_with_mail: This change will update `Bamboo.Mailer` `__using__` macro to allow passing
of optional `config:` keyword to the `build_config/1` function. `build_config/1`
now pattern matches on `config:`, and will pass the value to
`Bamboo.Mailer.build_config/3`. This function will then merge the
dynamic configuration with the default configuration.

:octocat: See #470

### DIFF
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -281,8 +281,6 @@ defmodule Bamboo.Mailer do
   end
 
   def build_config(mailer, otp_app, optional_overrides \\ %{}) do
-    optional_overrides = Enum.into(optional_overrides, %{})
-
     otp_app
     |> Application.fetch_env!(mailer)
     |> Map.new()
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -77,13 +77,14 @@ defmodule Bamboo.Mailer do
 
       otp_app = Keyword.fetch!(opts, :otp_app)
 
-      defp build_config([config: dynamic_config_overrides]) do
+      defp build_config(config: dynamic_config_overrides) do
         Bamboo.Mailer.build_config(
           __MODULE__,
           unquote(otp_app),
           dynamic_config_overrides
         )
       end
+
       defp build_config(_) do
         Bamboo.Mailer.build_config(__MODULE__, unquote(otp_app))
       end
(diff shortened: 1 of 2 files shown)


test/lib/bamboo/mailer_test.exs
```diff
@@ -253,6 +253,36 @@ defmodule Bamboo.MailerTest do
...
+
+    test "deliver_later/2 overrides Adapter config with the 'config:' option" do
+      email = new_email(to: "baz@qux.com")
+
+      override_config = %{
+        foo: :qux,
+        something: :groovy
+      }
+
+      Mailer.deliver_now(email, config: override_config)
```
maymillerricci: Was this meant to be `deliver_later`?

lib/bamboo/mailer.ex
```diff
@@ -264,10 +280,13 @@ defmodule Bamboo.Mailer do
     build_config(mailer, otp_app)
   end
 
-  def build_config(mailer, otp_app) do
+  def build_config(mailer, otp_app, optional_overrides \\ %{}) do
+    optional_overrides = Enum.into(optional_overrides, %{})
```
maymillerricci: Is the idea here that people might be passing things in as optional overrides that are in different formats?

test/lib/bamboo/mailer_test.exs
stratigos: :eagle: Thanks so much for catching that! 

Testing the expected test subject with a41e4d74de15e758793a8669bc75821e1f7c0a60 :sweat_smile:

lib/bamboo/mailer.ex
stratigos: :thinking: Im trying to recall where my head was when I made this decision.

Im thinking I had a hard time determining if things were passed in as Keyword Lists or Maps, or both. I think I saw this technique in a test, though that scenario wouldnt apply here.

From my current perspective, it feels like defensive programming. I intend to remove it now.

stratigos: Defensively casting the argument as a Map is removed with 8b1abd0f025ffe99a4c9008db2829f05dc0c1b86 :world_map:

gfviegas:
I was looking for this feature but it's not documented, at least not quite as easy as I could find. 

Going to PR it as soon as I have a break tho

PR #498, opened 2019-09-30
Pass opts to hackney

mtarnovan:
This allows passing down opts to hackney. We have need of this as we want to increase the default timeout.


mtarnovan:
Thanks for the release, May!

PR #519, opened 2020-01-31
Support SendGrid send_at field

andrewjensen:
This adds support for SendGrid's send_at field, according to SendGrid's [guide on scheduling](https://sendgrid.com/docs/API_Reference/SMTP_API/scheduling_parameters.html) and [API reference](https://sendgrid.com/docs/api-reference).

This is useful for scenarios when you need to do a lot of computational work to send a batch of emails, but you want them all delivered at the same time.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ Mandrill offers extra features on top of regular SMTP email like tagging, merge
 vars, templates, and scheduling emails to send in the future. See
 [`Bamboo.MandrillHelper`].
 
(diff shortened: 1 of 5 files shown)


andrewjensen:
@maymillerricci @cpytel this has been open for a couple weeks now. Is anyone available to review it? I think it's ready to go, since it includes test coverage, typespecs, and doc comments. Let me know if you need anything else from my end.

PR #521, opened 2020-03-06
Fix mailgun adapter error reporting

antondomratchev:
Mailgun adapter was raising errors with encoded body.  This caused the
`api_error` handler to report one large string as the param. This was hard
to debug and know which email was throwing the error because
`printable_limit` config of `inspect\2` was truncating the string to 4096 
characters [See here](https://hexdocs.pm/elixir/1.8.1/Inspect.Opts.html). Due to the alphabetical sorting of map keys we only got
to see `from=` and `html=` params which didn't contain enough information to
know which email is failing. This change converts the body of mailgun request
back to a map so we can see all params for the failed request.

The limit of 150 keys is still being applied in case body has a lot of keys in it.

Curiously mailgun adapter is the only one that was raising the api_error with a string while other adapters used json decoded maps.


maymillerricci:
Thanks for this @antondomratchev! Do you think there's a way to add/update tests for this to make sure the functionality is working as expected?

antondomratchev:
@maymillerricci Sorry it took so long to get this shored up. I hope the test coverage is satisfactory.

antondomratchev:
bump

mix.lock
```diff
@@ -1,30 +1,30 @@
...
-  "plug": {:hex, :plug, "1.6.0", "90d338a44c8cd762c32d3ea324f6728445c6145b51895403854b77f1536f1617", [:mix], [{:cowboy, "~> 1.0.1 or ~> 1.1 or ~> 2.4", [hex: :cowboy, repo: "hexpm", optional: true]}, {:mime, "~> 1.0", [hex: :mime, repo: "hexpm", optional: false]}], "hexpm"},
-  "poison": {:hex, :poison, "3.1.0", "d9eb636610e096f86f25d9a46f35a9facac35609a7591b3be3326e99a0484665", [:mix], [], "hexpm"},
-  "ranch": {:hex, :ranch, "1.3.2", "e4965a144dc9fbe70e5c077c65e73c57165416a901bd02ea899cfd95aa890986", [:rebar3], [], "hexpm"},
-  "ssl_verify_fun": {:hex, :ssl_verify_fun, "1.1.1", "28a4d65b7f59893bc2c7de786dec1e1555bd742d336043fe644ae956c3497fbe", [:make, :rebar], [], "hexpm"},
-  "unicode_util_compat": {:hex, :unicode_util_compat, "0.3.1", "a1f612a7b512638634a603c8f401892afbf99b8ce93a45041f8aaca99cadb85e", [:rebar3], [], "hexpm"},
+  "certifi": {:hex, :certifi, "2.3.1", "d0f424232390bf47d82da8478022301c561cf6445b5b5fb6a84d49a9e76d2639", [:rebar3], [{:parse_trans, "3.2.0", [hex: :parse_trans, repo: "hexpm", optional: false]}], "hexpm", "e12d667d042c11d130594bae2b0097e63836fe8b1e6d6b2cc48f8bb7a2cf7d68"},
```
germsvel, line 2: Did any dependencies change with this PR? I'm curious why we're seeing this.

germsvel: I was just doing some work on Bamboo, and I see these are coming from a simple `mix deps.get`. It seems we either missed committing the `mix.lock` file at some point or something else is going on. Either way, I don't think this is a blocker since we have to commit this file anyway.

lib/bamboo/adapters/mailgun_adapter.ex
```diff
@@ -267,4 +268,10 @@ defmodule Bamboo.MailgunAdapter do
   end
 
   defp encode_body(body_without_attachments), do: Plug.Conn.Query.encode(body_without_attachments)
+
+  defp decode_body({:multipart, _} = multipart_body), do: multipart_body
+
+  defp decode_body(body) when is_binary(body) do
+    URI.decode_query(body)
```
germsvel, line 275: I was giving this PR another look, and I realized this is something of a counterpart to the existing `encode_body/1` function defined in line 270, right? 

What do you think about using [`Plug.Conn.Query.decode`](https://hexdocs.pm/plug/Plug.Conn.Query.html#decode/4) since we use the `Plug.Conn.Query.encode` function to encode the data?

antondomratchev:
@germsvel Feel free to merge when ready. I don't seem to have a merge button available to me

germsvel:
Thanks for working on this @antondomratchev. 🎉

PR #523, opened 2020-03-12
Dynamic sendgrid key

### LINKED ISSUE #522: Advanced API key definitions for sendgrid.
It would be great to be able to do something like the way stripity_stripe does configurations.

https://hexdocs.pm/stripity_stripe/readme.html#configuration

spunkedy:
Fixes #522

### DIFF
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -43,7 +43,7 @@ defmodule Bamboo.SendGridAdapter do
   import Bamboo.ApiError
 
   def deliver(email, config) do
-    api_key = get_key(config)
+    api_key = _get_key(config)
     body = email |> to_sendgrid_body(config) |> Bamboo.json_library().encode!()
     url = [base_uri(), @send_message_path]
 
@@ -63,16 +63,22 @@ defmodule Bamboo.SendGridAdapter do
   @doc false
   def handle_config(config) do
     # build the api key - will raise if there are errors
-    Map.merge(config, %{api_key: get_key(config)})
+    Map.merge(config, %{api_key: _get_key(config)})
   end
 
   @doc false
   def supports_attachments?, do: true
 
-  defp get_key(config) do
+  @doc """
+    Private function prefixed with `_` for testability.
+    This handles getting the config key.
+  """
+  def _get_key(config) do
     api_key =
       case Map.get(config, :api_key) do
         {:system, var} -> System.get_env(var)
+        {module_name, method_name, args} -> apply(module_name, method_name, args)
+        fun when is_function(fun) -> fun.()
         key -> key
       end
 
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -22,7 +22,6 @@ defmodule Bamboo.SendGridAdapter do
         api_key: "my_api_key"
           # or {:system, "SENDGRID_API_KEY"},
           # or {ModuleName, :method_name, []}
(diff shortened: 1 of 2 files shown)


lib/bamboo/adapters/send_grid_adapter.ex
```diff
@@ -63,16 +68,22 @@ defmodule Bamboo.SendGridAdapter do
...
 
   @doc false
   def supports_attachments?, do: true
 
-  defp get_key(config) do
+  @doc """
+    Private function prefixed with `_` for testability.
+    This handles getting the config key.
+  """
+  def _get_key(config) do
```
maymillerricci, line 81: Can you explain more about this? Can this not be private with `defp`?

test/lib/bamboo/adapters/send_grid_adapter_test.exs
```diff
@@ -54,6 +64,11 @@ defmodule Bamboo.SendGridAdapterTest do
     end
   end
 
+  @doc """
+  sendgrid secret example dynamic
```
maymillerricci, line 68: What do you think about either expanding on this and explaining it more, or dropping it entirely if it's kind of clear what this is here for?

lib/bamboo/adapters/send_grid_adapter.ex
```diff
@@ -19,7 +19,11 @@ defmodule Bamboo.SendGridAdapter do
       # In config/config.exs, or config.prod.exs, etc.
       config :my_app, MyApp.Mailer,
         adapter: Bamboo.SendGridAdapter,
-        api_key: "my_api_key" # or {:system, "SENDGRID_API_KEY"},
+        api_key: "my_api_key"
+          # or {:system, "SENDGRID_API_KEY"},
+          # or {ModuleName, :method_name, []}
+          # or fn -> "a resolveable function" end
+          # or &ModuleName.method_name/0
```
fuelen, line 26: AFAIK it is not possible to prepare a release with anonymous functions in config

spunkedy: they have worked for me in the past, but that was using distillery, I haven't tested it with mix release. I can remove that suggestion too if you think it warrants it

lib/bamboo/adapters/send_grid_adapter.ex
spunkedy: @maymillerricci sorry I didn't see your comments.

A few reasons I suggest the `_function_name` pattern:

- it allows for testing requiring less scaffolding in the ExUnit to get full coverage
- using a non private function allows for `@doc` tags for language server for those that are wanting to contribute and not sure what it does.
- it is a pattern I have seen throughout the community for these reasons.

I don't mind removing it if that's what you would prefer in your code base.

test/lib/bamboo/adapters/send_grid_adapter_test.exs
spunkedy: The doc or the function? I will expand the doc

spunkedy: Expanded in the latest changeset

lib/bamboo/adapters/send_grid_adapter.ex
spunkedy: updated the docs to remove it, since it's definitely not a normal pattern and 

https://elixirforum.com/t/how-to-use-anonymous-functions-in-configuration-files/23069
https://elixirforum.com/t/anonymous-functions-in-config-why-does-it-work/22202

germsvel:
Thank you so much @spunkedy!! I really appreciate coming back and updating all of this!

PR #526, opened 2020-03-26
Downgrade jason and lock pessimistic version on subversion like phx

joshchernoff:
I think we should follow the same pattern as phoenix and postgrex in that they lock the pessimistic version of `jason` at the minor version not the patch version. I think we can follow the examples of how both phoenix and postgrex  currently resolve their jason dep. 

Otherwise, users may find their selves facing the issue I recently found my self when I wanted to have jason 1.2 on my local reps. 

```
Failed to use "jason" (version 1.2.0) because
  bamboo (version 1.4.0) requires ~> 1.1.0
  phoenix (version 1.4.16) requires ~> 1.0
  postgrex (version 0.15.3) requires ~> 1.0
  mix.lock specifies 1.2.0
```


README.md
```diff
@@ -424,7 +424,7 @@ to design, develop, and grow your product.
 
 Thanks to @mtwilliams for an early version of the `SendGridAdapter`.
 
-
+-
```
joshchernoff: not sure whats up with this.  🤷‍♂️

halostatue:
Libraries, in general, should use the minor version in pessimistic versioning, and use additional constraints if they require a minimum version (e.g., `~> 1.0 and >= 1.1.2`).

joshchernoff:
I'm just merely stating that if it's good enough for phoenix and postgrex it should be good enough for bamboo.

tomtaylor:
This has been resolved in #531, although it'd be good to get a new release @maymillerricci?

joshchernoff:
So If I understand this write, someone came long did the same work I did after I made this PR, it gets merged and this they still forgot to include the mix.lock 🤦

joshchernoff:
And then there is this. Why is no one reading the backlog before submitting PRs?
https://github.com/thoughtbot/bamboo/pull/530

maymillerricci:
Sorry about that! Though I think we're okay keeping the `mix.lock` as it is?

We'll do a new release within the next couple of days.

maymillerricci:
New release is up! https://hex.pm/packages/bamboo/1.5.0

PR #530, opened 2020-04-09
Reduce Jason version constraints

AndrewDryga:
Otherwise, everyone using Bamboo would have a bad time upgrading other deps that bump Jason to v1.2. Especially in umbrella apps where you need to override version in every single app.


tomtaylor:
This can be closed, as it landed in #531.

PR #551, opened 2020-09-19
feat: Enable inline attachments for mailgun.

astjohn:
This PR enables inline attachments for mailgun.


astjohn:
At the time of writing, mailgun sets the content-id of the attachment to the supplied filename. I have an open ticket with them to see if there is a way to override this. If not, it's not 100% ideal, but it is a working solution.

astjohn:
Official stance from mailgun is that they don't currently support specifying `content-id`. Although not ideal, this remains a working solution until the day they decide to support something like that.

germsvel:
Thanks for opening this PR @astjohn. 

Do you have any links to Mailgun documentation that I could reference to help me review this PR?

astjohn:
It's been a while. Updating our own branch and coming back around to this. Here's the mailgun docs:
https://documentation.mailgun.com/en/latest/api_reference.html#api-reference
https://documentation.mailgun.com/en/latest/api-sending.html#sending

If you'd like to verify what I went through with Mailgun, feel free to email them. As far as I know, their stance hasn't changed on the matter.

germsvel:
@astjohn, thanks so much for looping back and rebasing the PR to be up to date!

Would you mind adding a test for the MailgunAdapter so the changes are tested? 

I think once we have that, this is good to go. 👍

doomspork:
Hi @astjohn 👋 Now that BEAM is managing Bamboo I'm trying to work through all open PRs. Can you please let us know the status of this? Can you please add the test coverage requested by @germsvel?

If this isn't a change you're interested in continuing to move forward we can close the PR 👍

PR #556, opened 2020-10-15
Elixir 1.11 warnings

tomciopp:
Fixes warnings emitted by the compiler and updates supervisors to new methodology.


lib/bamboo.ex
```diff
@@ -31,11 +31,9 @@ defmodule Bamboo do
   end
 
   def start(_type, _args) do
-    import Supervisor.Spec
-
     children = [
-      worker(Bamboo.SentEmail, []),
-      supervisor(Task.Supervisor, [[name: Bamboo.TaskSupervisorStrategy.supervisor_name()]])
+      Bamboo.SentEmail,
+      {Task.Supervisor, [name: Bamboo.TaskSupervisorStrategy.supervisor_name()]}
```
germsvel, line 36: 🎉

lib/bamboo/adapters/local_adapter.ex
```diff
@@ -30,13 +30,12 @@ defmodule Bamboo.LocalAdapter do
...
   @behaviour Bamboo.Adapter
 
-  @doc "Adds email to `Bamboo.SentEmail` and opens it in new browser tab"
+  @doc "Adds email to `Bamboo.SentEmail`, can automatically open it in new browser tab"
   def deliver(email, %{open_email_in_browser_url: open_email_in_browser_url}) do
     %{private: %{local_adapter_id: local_adapter_id}} = SentEmail.push(email)
     open_url_in_browser("#{open_email_in_browser_url}/#{local_adapter_id}")
   end
 
-  @doc "Adds email to `Bamboo.SentEmail`"
```
germsvel, line 39: Add of curiosity, why remove the doc? Could we revert changes to this?

mix.exs
```diff
@@ -53,13 +54,14 @@ defmodule Bamboo.Mixfile do
...
       {:cowboy, "~> 1.0", only: [:test, :dev]},
-      {:phoenix, "~> 1.1", only: :test},
+      {:phoenix, "~> 1.1", optional: true},
       {:phoenix_html, "~> 2.2", only: :test},
-      {:excoveralls, "~> 0.4", only: :test},
-      {:floki, "~> 0.8", only: :test},
-      {:ex_doc, "~> 0.20", only: :dev},
+      {:excoveralls, "~> 0.13", only: :test},
+      {:floki, "~> 0.29", only: :test},
+      {:ex_doc, "~> 0.23", only: :dev},
```
germsvel, line 64: Could you provide information as to why we needed these changes? Did we have to update dependencies?

germsvel (commented):
Thanks for opening a PR trying to fix Elixir 1.11 warnings. 

I had a few questions, and could you provide more information for the changes you're making? I haven't run through Bamboo with Elixir 1.11, so I'm not 100% sure what warnings the changes correspond to.

lib/bamboo/adapters/local_adapter.ex
tomciopp: Elixir 1.11 throws a warning for this behavior. You can't add two different calls to `@doc` for the same function, you have to condense it into a single place.

mix.exs
tomciopp: Force of habit, since all of these deps are limited to :dev and :test there should not be any issues with updating.

tomciopp: We add `:mime` since it is used in `lib/bamboo/attachment.ex` and not directly referenced.

tomciopp: We mark phoenix as optional since it is used in `lib/bamboo/phoenix.ex` with `Phoenix.View.render_to_string`. It's not 100% necessary to use that module but it's very likely to be used.

rzane:
Hi @germsvel, would it be possible to get this merged?

mix.exs
```diff
@@ -21,7 +21,8 @@ defmodule Bamboo.Mixfile do
       start_permanent: Mix.env() == :prod,
       package: package(),
       docs: [main: "readme", extras: ["README.md"]],
-      deps: deps()
+      deps: deps(),
+      xref: [exclude: [IEx, ExMachina]]
```
germsvel, line 25: Hi @tomciopp, 

Thanks so much for all the extra information you gave regarding the dependencies. I was looking through this PR again, and this caught my eye. Could you give some extra info as to why we're excluding these from xref? 

When I read [the Elixir 1.11 announcement](https://elixir-lang.org/blog/2020/10/06/elixir-v1-11-0-released/#compiler-checks-application-boundaries), it seems like these two should not be excluded here. 

It seems like `iex` should fall under the "part of Elixir" portion of libraries, so we should put it in `extra_applications`, right? Or to dig even further, do we even need to list `IEx` at all? Where does Bamboo use `IEx`? 

And `ExMachina` is defined in `deps`, so it seems like we shouldn't need to list it here either, right?

Thanks for all the info on the rest of the PR. I think this is the last question I have.

rzane: If I'm not mistaken, adding `:iex` to `:extra_applications` would imply that IEx should be started whenever Bamboo starts, which I don't think is the desired behavior.

As for ExMachina, I'm not sure why that needed to be excluded.

tomciopp: IEx is used in the following file
`lib/mix/start_sent_email_viewer_task.ex:70`

Here is the warning emitted

```
warning: IEx.started?/0 defined in application :iex is used by the current application but the current application does not directly depend on :iex. To fix this, you must do one of:

  1. If :iex is part of Erlang/Elixir, you must include it under :extra_applications inside "def application" in your mix.exs

  2. If :iex is a dependency, make sure it is listed under "def deps" in your mix.exs

  3. In case you don't want to add a requirement to :iex, you may optionally skip this warning by adding [xref: [exclude: IEx]] to your "def project" in mix.exs
```

We don't want IEx to be started when Bamboo starts so we use option 3 to exclude it.

tomciopp: There is a separate warning with ExMachina, but I believe it is related to the library itself. I will remove the module from the xref exclude list. I've reproduced the warning below for your convenience.

```
warning: incompatible types:

    map() !~ Bamboo.NormalizeAndPushStrategy

in expression:

    # test/support/normalize_and_push_strategy.ex:2
    custom_strategy_module.function_name

where "custom_strategy_module" was given the type Bamboo.NormalizeAndPushStrategy in:

    # test/support/normalize_and_push_strategy.ex:2
    custom_strategy_module = Bamboo.NormalizeAndPushStrategy

where "custom_strategy_module" was given the type map() (due to calling var.field) in:

    # test/support/normalize_and_push_strategy.ex:2
    custom_strategy_module.function_name

HINT: "var.field" (without parentheses) implies "var" is a map() while "var.fun()" (with parentheses) implies "var" is an atom()

Conflict found at
  test/support/normalize_and_push_strategy.ex:2: Bamboo.NormalizeAndPushStrategy.__using__/1
```

germsvel (approved):
Thanks @tomciopp for this PR and for providing all that helpful information. 🏅  

This looks good, so I'll go ahead and merge it.

PR #559, opened 2020-10-29
Remove `@email_view_module` from Bamboo.Phoenix

germsvel:
What changed?
============

We remove the unnecessary setting of the `view` provided to `Bamboo.Phoenix` as `@email_view_module`. By setting it to a module attribute, we create a compile-time dependency, when we can simply pass it into the `render_email` function. It was originally set in [this commit](https://github.com/thoughtbot/bamboo/commit/9e08553595e8def2fa0cccb2662583cd21fd0f55#diff-dd737ef2fe5a0cd4fa37a32d32dc27c7d861e337fef92bae9e1c48cb8e1871ea).


brimatteng (approved):
👍

PR #563, opened 2020-11-13
Add assign_many/2

andersonmcook:
This will add the ability for a map or keyword to be used when making assigns for an email.

It also adds tests for `assign/3` and `assign_many/2`

### DIFF
--- a/test/lib/bamboo/phoenix_test.exs
+++ b/test/lib/bamboo/phoenix_test.exs
@@ -53,6 +53,17 @@ defmodule Bamboo.PhoenixTest do
       new_email()
       |> render("template.foobar")
     end
+
+    def email_with_assigns do
+      new_email()
+      |> assign(:a, "a")
+      |> assign(:b, "b")
+      |> assign(:c, "c")
+    end
+
+    def email_with_many_assigns do
+      assign_many(new_email(), a: "a", b: "b", c: "c")
+    end
   end
 
   test "render/2 allows setting a custom layout" do
@@ -113,5 +124,14 @@ defmodule Bamboo.PhoenixTest do
     assert_raise RuntimeError, ~r/documentation only/, fn ->
       Bamboo.Phoenix.render(:foo, :foo, :foo)
     end
+  end
+
+  test "assign/3 will assign a variable for the email" do
+    assert %{assigns: %{a: "a", b: "b", c: "c"}} = Email.email_with_assigns()
+  end
+
+  test "assign_many/2 will assign many variables for the email" do
+    assert %{assigns: %{a: "a", b: "b", c: "c"}} =Email.email_with_many_assigns()
+    assert Email.email_with_assigns() == Email.email_with_many_assigns()
   end
 end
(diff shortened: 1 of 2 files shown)


test/lib/bamboo/phoenix_test.exs
```diff
@@ -114,4 +125,13 @@ defmodule Bamboo.PhoenixTest do
       Bamboo.Phoenix.render(:foo, :foo, :foo)
     end
   end
+
+  test "assign/3 will assign a variable for the email" do
+    assert %{assigns: %{a: "a", b: "b", c: "c"}} = Email.email_with_assigns()
+  end
+
+  test "assign_many/2 will assign many variables for the email" do
+    assert %{assigns: %{a: "a", b: "b", c: "c"}} =Email.email_with_many_assigns()
```
kyleVsteger, line 134: format?

andersonmcook: Fixed!  Thanks!

germsvel (commented):
Thanks for opening this PR and contributing to Bamboo. 

I'd love to know more about the use case you have for `assign_many/2`. 

I realize it must be coming from the desire not to set one `assign` at a time. But I was curious, would using [`render/3`](https://hexdocs.pm/bamboo/Bamboo.Phoenix.html#render/3) do what you want here? 

You can pass the assigns as a map as the last argument: 

```elixir
# this pr suggestion
new_email()
|> assign_many(a: "a", b: "b")
|> render(:welcome)

# with render/3
new_email()
|> render(:welcome, a: "a", b: "b")
```

andersonmcook:
> Thanks for opening this PR and contributing to Bamboo.
> 
> I'd love to know more about the use case you have for `assign_many/2`.
> 
> I realize it must be coming from the desire not to set one `assign` at a time. But I was curious, would using [`render/3`](https://hexdocs.pm/bamboo/Bamboo.Phoenix.html#render/3) do what you want here?
> 
> You can pass the assigns as a map as the last argument:
> 
> ```elixir
> # this pr suggestion
> new_email()
> |> assign_many(a: "a", b: "b")
> |> render(:welcome)
> 
> # with render/3
> new_email()
> |> render(:welcome, a: "a", b: "b")
> ```

Using `render` may indeed solve the problem!  I no longer work on the codebase where I was using Bamboo, but the impetus was from seeing multiple `assign`s piped into one another and feeling like there could be a friendlier way to take a map or keyword, much in the way you can use multiple `Map.put`s or one `Map.merge`.

germsvel:
Thanks for letting me know @andersonmcook!

germsvel:
I'm sorry to have to close this PR since `Bamboo.Phoenix` was extracted to [bamboo_phoenix]. 

If someone is interested in continuing this work, I think we can either:

- Open the equivalent PR in [bamboo_phoenix], if we want to add it to `Bamboo.Phoenix`, or 
- If we want to add multiple assigns for [`Bamboo.Template`](https://hexdocs.pm/bamboo/Bamboo.Template.html#assign/3), then open a PR in this repo.

Regardless of which PR is opened, I think it'd be nice to overload the `assign` function by introducing an `assign/2` which would take the `map` version instead of a `key` and `value` (similar to how [Phoenix.LiveView.assign/2](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#assign/2) works). 

[bamboo_phoenix]: https://github.com/thoughtbot/bamboo_phoenix

PR #566, opened 2020-11-30
Add support for Mailgun recipient variables

hfjallemark:
Adds support for Mailgun recipient variables:
https://documentation.mailgun.com/en/latest/user_manual.html#recipient-variables


germsvel (approved):
This looks great. 🎉   Thanks for contributing and for adding a link to the documentation for reference 🙏

PR #567, opened 2020-12-01
Improve attachment support detection

sasa1977:
This fixes a subtle bug which we encountered on one project in tests.

The problem is that `function_exported?` returns false if the module is not loaded (see [here](https://hexdocs.pm/elixir/Kernel.html?#function_exported?/3)). Consequently, if the beam is started in interactive mode (default when mix is used), Bamboo may incorrectly conclude that attachments are not supported.

This can lead to some strange behaviour in tests. For example, we had a test failing if executed in isolation, but sometimes the test would pass if all the tests are running. The test would always pass if the test adapter is recompiled while invoking `mix test`.

Note that the same issue can occur in prod if OTP release is not used.

### DIFF
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -213,7 +213,8 @@ defmodule Bamboo.Mailer do
   defp validate_attachment_support(%{attachments: []} = email, _adapter), do: email
 
   defp validate_attachment_support(email, adapter) do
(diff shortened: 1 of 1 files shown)


germsvel (approved):
Thanks for catching this error and opening a PR @sasa1977 . 

Looks great 🎉

Issue #568, opened 2020-12-09
Error in sent_email_viewer -> index.html.eex if more 1 sender was added to from


esurat20:
I got an error when i was trying to add several senders to `from`
`no function clause matching in Bamboo.SentEmailViewerPlug.Helper.format_email_address/1`
```
def format_text_body(body) do
    HTML.html_escape(body || "")
end

def format_email_address({nil, address}), do: address

def format_email_address({name, address}) do
    "#{name} &lt;#{address}&gt;"
end
```

senders map: `[{"Test 1", "test1@mail.com"}, {"test2", "test2@mail.com"}]`

Need to add `format_email_address` variant which can handle list.

I`ve fixed this problem but can not push my branch to the repo.

germsvel:
Thanks for pointing this out @esurat20. If you want to open a PR (I'd love to take a look at it), you have to fork the repo, push the branch to that repo, and then open a PR from that one to the `bamboo` repo. I don't think you can just push a new branch to the `bamboo` repo itself. Does that make sense?

esurat20:
@germsvel  Yep sure. I will create a fork.

esurat20:
@germsvel please check related PR.

esurat20:
@germsvel any updates?

germsvel:
Hi @esurat20, thanks for checking back on this. This slipped past me somehow. Just submitted a review. Thanks for contributing to Bamboo!

PR #569, opened 2020-12-09
feat: show all senders in sent email viewer template

esurat20:
[Issue 568](https://github.com/thoughtbot/bamboo/issues/568)

### DIFF
--- a/lib/bamboo/plug/sent_email_viewer/helper.ex
+++ b/lib/bamboo/plug/sent_email_viewer/helper.ex
@@ -42,6 +42,9 @@ defmodule Bamboo.SentEmailViewerPlug.Helper do
     HTML.html_escape(body || "")
   end
 
+  def format_email_address(emails) when is_list(emails),
+    do: Enum.map(emails, &format_email_address/1)
+
   def format_email_address({nil, address}), do: address
 
   def format_email_address({name, address}) do
--- a/test/lib/bamboo/plug/sent_email_viewer_plug_test.exs
+++ b/test/lib/bamboo/plug/sent_email_viewer_plug_test.exs
@@ -67,6 +67,28 @@ defmodule Bamboo.SentEmailViewerPlugTest do
     end
   end
 
+  test "shows all senders" do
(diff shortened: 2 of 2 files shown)


lib/bamboo/plug/sent_email_viewer/helper.ex
```diff
@@ -42,6 +42,9 @@ defmodule Bamboo.SentEmailViewerPlug.Helper do
     HTML.html_escape(body || "")
   end
 
+  def format_email_address(emails) when is_list(emails),
```
germsvel, line 45: I wonder if there's a better way to handle multiple emails. My hesitation with this approach is that this helper function was meant for a single email. Bamboo uses an internal format for addresses (a tuple of `{name, address}`), so this function is meant to take that and display the email nicely. 

Is it possible to iterate over the email addresses elsewhere and format them one by one with the `format_email_addess/1` below as it stands?

germsvel (commented):
Thanks for opening this PR. I wasn't aware that we could have multiple senders in an email. That's really interesting! 

I would like to find a better approach than overriding the `format_email_address/1` helper function. I'll try to dig deeper when I have some time to see if I can offer a suggested approach. But if you have time before I do that, and you find a different approach, I'd love to hear it.

lib/bamboo/plug/sent_email_viewer/helper.ex
germsvel: After looking into this a bit more, I think this is a good way to handle this. We could iterate on the `sent_email_viewer/index.html` template, if `from` is a list of emails, but I think I like this better than that approach. 

It's also similar to how we the `Bamboo.Formatter` handles lists, though in that case we're generating the 2-tuples rather than formatting for presenting in the email. Still, makes me feel like this is consistent with how we've dealt with this in the past.

germsvel (approved):
Finally had time to go back and look into this a bit more. I think this approach looks good. 

Thanks for contributing to Bamboo! 🎉

Issue #505, opened 2019-12-15
Quick thoughts on returning {:ok, response} | {:error, response}


Ivor:
Hi :) It's been interesting reading through the evolution of the `response: true` feature. 

I want to match on the response from `deliver_now` to determine whether the email was successfully sent or not. Looking at the tests in `mailer_test` I see that your stubbed response contains this `%{status_code: status, headers: headers, body: response}` which seems to be what the Mandrill and SendGrid adapters also return, but only if the status code is < 299

(could formalise that as a struct that the implementation of the adapter behaviour needs to return -> %Bamboo.Response{})

The other status codes for that adapter raise an exception, so the `response: true` option will only return a response when the request was successful. If the implementations of the adapters returned `{:error, %{status_code: status, headers: headers, body: response}` instead of `raise_api_exception` then it would be possible to match on that. 

Renaming `deliver_now` to `deliver_now!` with the current behaviour of raising exceptions and then perhaps changing `deliver_now` to return `{:ok, %Response{}} | {:error, %Response{}}`

For the initial need to handle email delivery failures -> can we assume that if the email was not successfully delivered an exception will be raised? Is there somewhere where the expected range of exceptions is documented? 

I don't know very much about emails and email delivery and I am probably not understanding al of the issues involved and apologise if my comments are short-sighted. I also suspect that there are perhaps changes planned for a future breaking version update that could include this. 

If someone can perhaps document what the expected behaviour is for failures more explicitly it will already help a lot. Thanks for all the great work on this. Bamboo is absolutely great!

joshchernoff:
> Renaming deliver_now to deliver_now! with the current behaviour of raising exceptions and then 
> perhaps changing deliver_now to return {:ok, %Response{}} | {:error, %Response{}}

I 1+ this. Any beyond having deliver_now return a tuple I think any method whos default is to raise if not complete really should be named with a bang.

germsvel:
Thanks for opening this issue @Ivor. I really like the idea of returning an ok/error tuple for `deliver_now` and adding a `!` for the raising errors. As I'm thinking this, I wonder if it makes sense to use ok/error tuples in regular scenarios and let `response: true` still be about the `response` that comes from the adapter. So we could change the api to something like this: 

```elixir
# success deliver_now
{:ok, email} = deliver_now(email)
{:ok, email, response} = deliver_now(email, response: true)

# error deliver_now
{:error, email} = deliver_now(email)
[...]

Issue #403, opened 2018-06-16
How to handle email delivery failures


denvaar:
I am using Bamboo to send account confirmation emails when someone creates an account. I'm using `deliver_later` to send emails in the background. In my application, an account is initially created as "inactive" until they use the confirmation link in the email.

I want to be able to handle the case where a fake email is used to create an account, because I don't want a bunch of fake stuff hanging around in the database. My application doesn't know it's fake until this exception is raised from within a different process:

`(Bamboo.SMTPAdapter.SMTPError) There was a problem sending the email through SMTP`

I looked at the docs and came up with a solution using a custom delivery strategy, but I am not sure if this type of thing is what it's intended for, so I wanted to get some validation, and check if this is a good thing to do. As you can see from my code below, I am basically just catching the exception and then deleting the fake user account from the database.

```elixir
defmodule MyApp.AsyncEmailStrategey do
  @behaviour Bamboo.DeliverLaterStrategy

  alias MyApp.Accounts

  def deliver_later(adapter, email, config) do
    Task.async fn ->
      try do
        adapter.deliver(email, config)
      rescue
        error in Bamboo.SMTPAdapter.SMTPError ->
          remove_user_account(email)
          raise error
      end
    end
  end

  defp remove_user_account(email) do
    with [nil: bad_email] = email.to,
         user = Accounts.get_user_by_email(bad_email)
    do
      Accounts.delete_user(user)
    end
  end
end
```

claytongentry:
Would like to bump this issue. 

We have tested a similar try / rescue flow using [Amazon SES Simulator inbox addresses](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/mailbox-simulator.html) and have found the adapter does not raise as we'd expect on deliveries to the faulty addresses, e.g. `bounce@simulator.amazonses.com`. 

This makes it difficult to clean our lists of bouncing, suppressed or complaint emails.

claytongentry:
Hey @denvaar, don't know if you're still having this issue but the solution we ended up going with is subscribing to bounce and complaint notifications via SNS. 

https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notifications-via-sns.html

blacksph3re:
Still it would be nice having this for any smtp service which doesn't support separate bounce/complaint notifications

germsvel:
[...]

PR #571, opened 2020-12-28
Add Mailer.deliver_now! and deliver_later!

germsvel:
Closes: https://github.com/thoughtbot/bamboo/issues/355, https://github.com/thoughtbot/bamboo/issues/409, https://github.com/thoughtbot/bamboo/issues/505

Possibly helps: https://github.com/thoughtbot/bamboo/issues/403, https://github.com/thoughtbot/bamboo/issues/346

Supersedes PR: https://github.com/thoughtbot/bamboo/pull/393

What changed?
=============

We update `deliver_now` and `deliver_later` to not raise on errors. They now return an `{:ok, email}` (or `{:ok, email, response}` with `response: true`) or `{:error, error}`.

We also introduce `deliver_now!` and `deliver_later!` for those who liked the previous behavior. Thus, for those who don't want to change the behavior of Bamboo, it should be possible to simply replace what used to be `deliver_now` and `deliver_later` with the `!` equivalents.

Those who want to handle errors can receive ok/error tuples with `deliver_now` and `deliver_later`.

Note on `deliver_later` and `deliver_later!`

`deliver_later` now returns an `{:ok, email_to_send}` or `{:error, error}` if there is an error in the email validations _before_ we schedule the email sending.

If you want to keep the existing behavior, use `deliver_later!`. `deliver_later!` will raise an error if we have email validation failures before we schedule the email.

Regardless of whether you choose `deliver_later` or `deliver_later!`, `TaskSupervisorStrategy` raises an error because the delivery happens in a separate process. Thus the behavior of the `TaskSupervisorStrategy` does not change with this commit.

Note on empty recipients (not nil)
-------------------------------

While working on this update, I noticed we do not raise an error when `to, cc, bcc` are all empty lists (after formatting). We decide to keep in line with this behavior, so we do not return an error in those cases either.

We do not want to raise on scenarios the recipient list is empty because that means the user of bamboo interacted with those lists. If all of those fields are `nil`, we do return an error or raise an error because it means the user of bamboo might not have set anything accidentally (`nil` is the default value for all of those). That has always been the case, and will continue to be.

`Bamboo.ApiError.build_api_error` for adapters
------------------------------------

We also add a `Bamboo.ApiError.build_api_error` function that is the counterpart to the `raise_api_error`. That way, adapters can use the `build_api_error` function to return `{:error, %Bamboo.ApiError{}}` tuples instead of raising errors.

Adapter-specific PRs: 

- [x] SendGrid https://github.com/thoughtbot/bamboo/pull/572
- [x] Mailgun https://github.com/thoughtbot/bamboo/pull/573
- [x] Mandrill https://github.com/thoughtbot/bamboo/pull/574

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ email in fitting places within your application.
 defmodule MyApp.SomeControllerPerhaps do
   def send_welcome_email do
     Email.welcome_email()   # Create your email
-    |> Mailer.deliver_now() # Send your email
+    |> Mailer.deliver_now!() # Send your email
   end
 end
 ```
@@ -200,9 +200,10 @@ process completion (e.g. a web request in Phoenix). Bamboo provides a
 also provides a [`Bamboo.DeliverLaterStrategy`] behaviour that you can
 implement to tailor your background email sending.
 
-By default, `deliver_later`uses [`Bamboo.TaskSupervisorStrategy`]. This strategy
-sends the email right away, but it does so in the background without linking to
-the calling process, so errors in the mailer won't bring down your app.
+By default, `deliver_later` uses [`Bamboo.TaskSupervisorStrategy`]. This
+strategy sends the email right away, but it does so in the background without
+linking to the calling process, so errors in the mailer won't bring down your
+app.
 
 You can also create custom strategies by implementing the
 [`Bamboo.DeliverLaterStrategy`] behaviour. For example, you could create
--- a/lib/bamboo/adapter.ex
+++ b/lib/bamboo/adapter.ex
@@ -35,7 +35,9 @@ defmodule Bamboo.Adapter do
       end
   """
 
-  @callback deliver(%Bamboo.Email{}, %{}) :: any
+  @type error :: Exception.t() | String.t()
+
+  @callback deliver(%Bamboo.Email{}, %{}) :: {:ok, any} | {:error, error}
   @callback handle_config(map) :: map
   @callback supports_attachments? :: boolean
 end
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -2,8 +2,8 @@ defmodule Bamboo.Mailer do
   @moduledoc """
   Functions for delivering emails using adapters and delivery strategies.
 
-  Adds `deliver_now/1` and `deliver_later/1` functions to the mailer module it
-  is used by.
+  Adds `deliver_now/1`, `deliver_now!/1`, `deliver_later/1` and
+  `deliver_later!/1` functions to the mailer module in which it is used.
 
   Bamboo [ships with several adapters][available-adapters]. It is also possible
   to create your own adapter.
@@ -49,7 +49,7 @@ defmodule Bamboo.Mailer do
         end
       end
 
-  You are now able to send emails with your mailer module where you sit fit
+  You are now able to send emails with your mailer module where you see fit
   within your application.
   """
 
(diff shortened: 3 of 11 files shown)


README.md
```diff
@@ -145,7 +145,7 @@ email in fitting places within your application.
 defmodule MyApp.SomeControllerPerhaps do
   def send_welcome_email do
     Email.welcome_email()   # Create your email
-    |> Mailer.deliver_now() # Send your email
+    |> Mailer.deliver_now!() # Send your email
```
lady3bean, line 148: Do you think we should include documentation of `deliver_now` as well in the README?

lib/bamboo/adapter.ex
```diff
@@ -35,7 +35,9 @@ defmodule Bamboo.Adapter do
       end
   """
 
-  @callback deliver(%Bamboo.Email{}, %{}) :: any
+  @type error :: Exception.t() | String.t()
```
lady3bean, line 38: 🎉

lib/bamboo/mailer.ex
```diff
@@ -132,51 +168,91 @@ defmodule Bamboo.Mailer do
...
-      debug_unsent(email)
-    else
-      debug_sent(email, adapter)
-      config.deliver_later_strategy.deliver_later(adapter, email, config)
+      {:ok, email}
+    end
+  end
+
+  @doc false
+  def deliver_later!(adapter, email, config) do
```
lady3bean, line 247: I like how the functions with the bangs build on their plain counterparts here. Makes what is changing in this PR clear, too.

lib/bamboo/mailer.ex
```diff
@@ -200,45 +276,49 @@ defmodule Bamboo.Mailer do
...
   end
 
   defp validate_from_address(%{from: nil}) do
-    raise Bamboo.EmptyFromAddressError, nil
+    {:error, %Bamboo.EmptyFromAddressError{}}
   end
 
   defp validate_from_address(%{from: {_, nil}}) do
-    raise Bamboo.EmptyFromAddressError, nil
+    {:error, %Bamboo.EmptyFromAddressError{}}
```
lady3bean, line 309: I like that these changes result in more idiomatic elixir, returning the error tuple

lady3bean (approved):
This looks great! The tests look good and made me feel more confident that these changes make sense.

Do you think that since this is a breaking change, there should be more documentation or process when it comes to releasing this?

README.md
germsvel: Yeah, maybe it'd be good to add a little more docs in the README about using `deliver_now` vs `deliver_now!`. I'll try to take a stab at that.

germsvel:
@lady3bean 

> Do you think that since this is a breaking change, there should be more documentation or process when it comes to releasing this?

Yes, the plan is to do a major bump (to 2.0). Hopefully that causes people to check the changelog. My plan right now (though happy to hear alternatives) was to write a little migration guide from 1.x to 2.0, include that guide in hex docs (just like we include the [readme](https://hexdocs.pm/bamboo/readme.html#content)), and link to that guide from the changelog. That way, the migration guide lives with the docs and is also discoverable from the changelog. What do you think?

lady3bean:
> @lady3bean
> 
> > Do you think that since this is a breaking change, there should be more documentation or process when it comes to releasing this?
> 
> Yes, the plan is to do a major bump (to 2.0). Hopefully that causes people to check the changelog. My plan right now (though happy to hear alternatives) was to write a little migration guide from 1.x to 2.0, include that guide in hex docs (just like we include the [readme](https://hexdocs.pm/bamboo/readme.html#content)), and link to that guide from the changelog. That way, the migration guide lives with the docs and is also discoverable from the changelog. What do you think?

That sounds great!

PR #573, opened 2020-12-28
Update Mailgun.deliver to return ok/error tuples

germsvel:

What changed?
==============

Updates Mailgin.deliver function to return ok/error tuples to abide by the new api.

Note on raising configuration errors
-----------------------------------

We do not change the raising of errors when they deal with configuration options. Those errors are categorically different. We want to return ok/error tuples when there's an error building/delivering email. If a user of Bamboo is missing a configuration variable, it seems important to raise an error.

### DIFF
--- a/lib/bamboo/adapters/mailgun_adapter.ex
+++ b/lib/bamboo/adapters/mailgun_adapter.ex
@@ -104,13 +104,13 @@ defmodule Bamboo.MailgunAdapter do
          ) do
       {:ok, status, _headers, response} when status > 299 ->
         body = decode_body(body)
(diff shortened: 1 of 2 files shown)


lady3bean:
This one is also failing on a missing `build_api_error` function, as is https://github.com/thoughtbot/bamboo/pull/572. Otherwise makes sense building off the other PR for deliver_now! and deliver_later!

```
== Compilation error in file lib/bamboo/adapters/mailgun_adapter.ex ==
** (CompileError) lib/bamboo/adapters/mailgun_adapter.ex:107: undefined function build_api_error/3
```

PR #574, opened 2020-12-28
Update Mandrill to return ok/error tuples

germsvel:

What changed?
============

Update MandrillAdapter to return an ok/error tuple to abide by the new behaviour api.

Note on configuration errors
--------------------------

We no longer raise api errors, but we do raise configuration errors. Since configuration errors do not fall under the category of email building/delivery errors, it seems fine that we raise those errors.


lady3bean:
also failing on undefined `build_api_error` function, like https://github.com/thoughtbot/bamboo/pull/572 and https://github.com/thoughtbot/bamboo/pull/573. But otherwise looks good.

```
== Compilation error in file lib/bamboo/adapters/mandrill_adapter.ex ==
** (CompileError) lib/bamboo/adapters/mandrill_adapter.ex:44: undefined function build_api_error/3
```

Issue #541, opened 2020-06-12
EEx rendering pipeline


lleger:
I have an umbrella app. The context app where the domain logic lives sends emails and the Phoenix app calls into the contexts and as a side effect sometimes emails are sent. 

Using Bamboo with `Bamboo.Phoenix` violates the app boundary, since it requires the templates to be in the Phoenix app, but the mailer logic in the context app. It also means the context app has an unnecessary dependency on Phoenix. It can also create circular dependencies. 

All of this works but it’s a bit of an anti-pattern, so I set out to fix it. I did this by writing an EEx rendering pipeline. I modeled the API off `Bamboo.Phoenix` so it was a one-line diff to switch over. The only difference is you don’t have access to any of the Phoenix view helpers, but that didn’t cause any issues for me. 

I’m wondering if the maintainers would see value in this being a part of Bamboo itself. If so I would be happy to open a PR for it! But I wanted to check before I did the work to do so, because I believe there may be some refactoring involved (e.g. the similar `assign` API can be hoisted and included in both `Bamboo.Phoenix` and a theoretical `Bamboo.EEx` instead of reimplementing the same code).

germsvel:
Hi @lleger, thanks so much for opening this issue! 

I've been trying to think about this for a little while, though I don't have a clear idea on whether or not this would be good to include in Bamboo. But I didn't want to leave this without a comment, because I have looked at it, and I'm thinking about it. 

I definitely get the desire to separate the email templates from the web layer that Phoenix provides. Currently, Phoenix [is not required](https://github.com/thoughtbot/bamboo/#using-phoenix-views-and-layouts) to use Bamboo, but I realize it provides a lot of niceties to generate html. So I imagine a lot of people feel like they have to depend on Phoenix. 

I do have a few questions: 

> All of this works but it’s a bit of an anti-pattern, so I set out to fix it. I did this by writing an EEx rendering pipeline. 

Is that EEx pipeline something we could see to get a better sense of what we're talking about?

> The only difference is you don’t have access to any of the Phoenix view helpers

Which helpers do you mean by that? You mean there wouldn't be any equivalents to `render` or route helpers, etc.? 

My last thought is, could this be a separate library to which we could point people if they don't want to depend on Phoenix? And are there any other EEX rendering libraries out there that already do something like this? 

Thanks again!

lleger:
Hi @germsvel! Sorry for the delayed response, you commented just about an hour before my daughter was born, so I'm just now seeing this!

[...]

Issue #323, opened 2017-10-04
Better view support


anthonator:
I'm curious if Bamboo has any plans on adding better view support. Currently, if you want to use views at all you need to piggyback off of Phoenix. This makes those of us not using Phoenix sad 🙁

Is there any interest in adding a generic view implementation to Bamboo?

germsvel:
Hi @anthonator, I realize this issue has been open for quite some time. Thanks for the patience! I'm starting to maintain Bamboo, and I just wanted to let you know that this work might happen. If you're interested, there's a discussion in https://github.com/thoughtbot/bamboo/issues/323 (I saw that issue before I saw this one).

germsvel:
Hi @anthonator, if this is something you're still interested in, I just opened a PR to introduce a view/template layer into Bamboo. I'd love to hear your thoughts. https://github.com/thoughtbot/bamboo/pull/576

germsvel:
Closing this because https://github.com/thoughtbot/bamboo/pull/576 is merged.

PR #576, opened 2020-12-31
Introduce Bamboo.Template and Bamboo.View

germsvel:
Closes: https://github.com/thoughtbot/bamboo/issues/541, https://github.com/thoughtbot/bamboo/issues/323

What changed?
=============

We introduce two new modules `Bamboo.Template` and `Bamboo.View` so that Bamboo users can use HTML and text templates without having to rely on Phoenix (not all Bamboo users use Phoenix).

The implementation is based heavily on Phoenix's Template and View modules. So a huge thanks to them for all the hard work.

Example usage:

```elixir
defmodule MyApp.EmailView do
  use Bamboo.View, path: "/path/to/this/view/templates"
end

defmodule MyApp.Email do
  use Bamboo.Template, view: MyApp.EmailView

  def welcome_email do
    new_email()
    |> put_layout({MyApp.EmailLayout, :email})
    |> render(:welcome_email)
  end
end
```

The goal is for `Bamboo.Template` to coexist with `Bamboo.Phoenix`, and `Bamboo.Phoenix` can be moved to a separate package in the future.

Difference from Phoenix Views
-----------------------------

When using `Bamboo.View` we have to specify the full path where the templates for that view will live (instead of the root path as is done in `Phoenix.View`).

We could do fancier things where we auto detect the location based on the module name, but that seems a little magical and perhaps unnecessary for now. We can always add that if we need it. For now, we can specify a view, and set the path for its templates.

Difference from Bamboo.Phoenix
-----------------------------

Unlike `Bamboo.Phoenix`, `Bamboo.Template` allows for overriding the view with `put_view/2` for cases when we don't have to rely on the view defined in `use Bamboo.Template, view: EmailView`.

Note on alternate implementation

The first pass at doing this used`EEx.function_from_file` instead `EEx.compile_file` and unquoting the compiled file as we define the function. But I ran into issues with the `function_from_file` approach because assigns needed to be keyword lists instead of maps. At that point, I looked at the approach `Phoenix.Template` uses, and decided to follow their approach of using `EEx.compile_file`.

See [Phoenix.Template's approach](https://github.com/phoenixframework/phoenix/blob/e05af0ad5527c9a192de122f3e43d7ed4c7a4c9f/lib/phoenix/template.ex#L346-L365)

Note on updated Bamboo.Phoenix tests
------------------------------------

Even though there were no changes to `Bamboo.Phoenix`, there were a couple of changes to `bamboo/phoenix_test`. That happened because we tried to reuse as many as the test templates as possible.

We did separate a `PhoenixLayoutView` since the layouts render their contents differently. `Bamboo.Phoenix` uses a version of Phoenix that still uses `render @view_module, @view_template, assigns` in the layout. But `Bamboo.View` is using `@inner_content`, like more recent versions of Phoenix.

We also have to include the `function_in_view/0` function in `bamboo/phoenix_test`'s `EmailView` since the view compiles a new template that calls that function.

### DIFF
--- a/lib/bamboo/template.ex
+++ b/lib/bamboo/template.ex
@@ -22,6 +22,20 @@ defmodule Bamboo.Template do
         end
       end
 
+  _You can optionally pass the html and text layouts when calling `use Bamboo.Template`_
+
+      defmodule MyApp.Email do
+        use Bamboo.Template,
+          view: MyApp.Email.AccountView
+          text_layout: {MyApp.Email.LayoutView, "email.text"},
+          html_layout: {MyApp.Email.LayoutView, "email.html"}
+
+        def welcome_email do
+          new_email()
+          |> render(:welcome) # Pass atom to render html AND plain text templates
+        end
+      end
+
   _Set both the text and HTML layout at the same time for an email._
 
       defmodule MyApp.Email do
@@ -64,7 +78,6 @@ defmodule Bamboo.Template do
 
         def welcome_email(user) do
           new_email()
-          |> put_html_layout({MyApp.Email.LayoutView, "email.html"})
           |> render(:welcome, user: user)
         end
       end
@@ -137,23 +150,26 @@ defmodule Bamboo.Template do
 
   import Bamboo.Email, only: [put_private: 3]
 
-  defmacro __using__(view: view_module) do
+  defmacro __using__(opts) do
+    view = Keyword.get(opts, :view, :unset)
+    html_layout = Keyword.get(opts, :html_layout, false)
+    text_layout = Keyword.get(opts, :text_layout, false)
+
     quote do
       import Bamboo.Email
       import Bamboo.Template, except: [render: 3]
 
       def render(email, template, assigns \\ []) do
-        Bamboo.View.render_email(unquote(view_module), email, template, assigns)
+        render_email(
+          unquote(view),
+          unquote(html_layout),
+          unquote(text_layout),
+          email,
+          template,
+          assigns
+        )
       end
     end
-  end
-
-  defmacro __using__(opts) do
-    raise ArgumentError, """
-    expected Bamboo.Template to have a view set, instead got: #{inspect(opts)}.
-
-    Please set a view e.g. use Bamboo.Template, view: MyEmailView
-    """
   end
 
   @doc """
@@ -267,5 +283,48 @@ defmodule Bamboo.Template do
   """
   def put_view(email, view) do
     email |> put_private(:view_module, view)
+  end
+
+  @doc false
+  def render_email(view, html_layout, text_layout, email, template, assigns) do
+    email
+    |> put_defaults(view, html_layout, text_layout)
+    |> merge_assigns(assigns)
+    |> render_template(template)
+  end
+
+  defp put_defaults(%{private: private} = email, view_module, html_layout, text_layout) do
+    private =
+      private
+      |> Map.put_new(:html_layout, html_layout)
+      |> Map.put_new(:text_layout, text_layout)
+      |> Map.put_new(:view_module, view_module)
+
+    %{email | private: private}
+  end
+
+  defp merge_assigns(%{assigns: email_assigns} = email, assigns) do
+    assigns = Map.merge(email_assigns, Enum.into(assigns, %{}))
+    Map.put(email, :assigns, assigns)
+  end
+
+  defp render_template(%{private: %{view_module: :unset}}, _template) do
+    raise ArgumentError, """
+    Expected Bamboo.Template to have a view set. Please set a view.
+
+    You can set it at the template level:
+
+      use Bamboo.Template, view: MyEmailView
+
+    or call it in your pipeline:
+
+      email
+      |> put_view(MyEmailView)
+      |> render("my_template.html")
+    """
+  end
+
+  defp render_template(email, template) do
+    Bamboo.View.render(email, template)
   end
 end
--- a/lib/bamboo/template.ex
+++ b/lib/bamboo/template.ex
@@ -124,7 +124,7 @@ defmodule Bamboo.Template do
 
       # lib/my_app/email/views/layout_view.ex
       defmodule MyApp.Email.LayoutView do
-        use Bamboo.View, path: "/lib/my_app/email/templates/layout"
+        use Bamboo.View, path: "lib/my_app/email/templates/layout"
       end
 
       # lib/my_app/email/templates/layout/email.html.eex
@@ -136,7 +136,7 @@ defmodule Bamboo.Template do
 
       # lib/my_app/email/views/account_view.ex
       defmodule MyApp.Email.AccountView do
-        use Bamboo.View, path: "/lib/my_app/email/templates/account"
+        use Bamboo.View, path: "lib/my_app/email/templates/account"
       end
 
       # lib/my_app/email/templates/account/welcome_email.html.eex
--- a/lib/bamboo/template.ex
+++ b/lib/bamboo/template.ex
@@ -1,0 +1,271 @@
+defmodule Bamboo.Template do
+  @moduledoc """
+  Render emails with layouts, view modules, and templates.
+
+  This module allows rendering emails with layouts and views. Pass an
+  atom (e.g. `:welcome_email`) as the template name to render both HTML and
+  plain text emails. Use a string if you only want to render one type, e.g.
+  `"welcome_email.text"` or `"welcome_email.html"`.
+
+  ## Examples
+
+  _Set the text and HTML layout for an email._
+
+      defmodule MyApp.Email do
+        use Bamboo.Template, view: MyApp.Email.AccountView
+
+        def welcome_email do
+          new_email()
+          |> put_text_layout({MyApp.Email.LayoutView, "email.text"})
+          |> put_html_layout({MyApp.Email.LayoutView, "email.html"})
+          |> render(:welcome) # Pass atom to render html AND plain text templates
+        end
+      end
+
+  _Set both the text and HTML layout at the same time for an email._
+
+      defmodule MyApp.Email do
+        use Bamboo.Template, view: MyApp.Email.AccountView
+
+        def welcome_email do
+          new_email()
+          |> put_layout({MyApp.Email.LayoutView, :email})
+          |> render(:welcome)
+        end
+      end
+
+  _Render both text and html emails without layouts._
+
+      defmodule MyApp.Email do
+        use Bamboo.Template, view: MyApp.Email.AccountView
+
+        def welcome_email do
+          new_email()
+          |> render(:welcome)
+        end
+      end
+
+  _Make assigns available to a template._
+
+      defmodule MyApp.Email do
+        use Bamboo.Template, view: MyApp.Email.AccountView
+
+        def welcome_email(user) do
+          new_email()
+          |> assign(:user, user)
+          |> render(:welcome)
+        end
+      end
+
+  _Make assigns available to a template during render call._
+
+      defmodule MyApp.Email do
+        use Bamboo.Template, view: MyApp.Email.AccountView
+
+        def welcome_email(user) do
+          new_email()
+          |> put_html_layout({MyApp.Email.LayoutView, "email.html"})
+          |> render(:welcome, user: user)
+        end
+      end
+
+  _Render an email by passing the template string to render._
+
+      defmodule MyApp.Email do
+        use Bamboo.Template, view: MyApp.Email.AccountView
+
+        def html_email do
+          new_email()
+          |> render("html_email.html")
+        end
+
+        def text_email do
+          new_email
+          |> render("text_email.text")
+        end
+      end
+
+  ## HTML Layout Example
+
+      # lib/my_app/email.ex
+      defmodule MyApp.Email do
+        use Bamboo.Template, view: MyApp.Email.AccountView
+
+        def welcome_email(person) do
+          base_email()
+          |> to(person)
+          |> subject("Welcome to MyApp")
+          |> assign(:person, person)
+          |> render(:welcome_email)
+        end
+
+        defp base_email do
+          new_email()
+          |> from("Rob Ot<robot@changelog.com>")
+          |> put_header("Reply-To", "editors@changelog.com")
+          # This will use the "email.html.eex" file as a layout when rendering html emails.
+          # Plain text emails will not use a layout unless you use `put_text_layout`
+          |> put_html_layout({MyApp.Email.LayoutView, "email.html"})
+        end
+      end
+
+      # lib/my_app/email/views/layout_view.ex
+      defmodule MyApp.Email.LayoutView do
+        use Bamboo.View, path: "/lib/my_app/email/templates/layout"
+      end
+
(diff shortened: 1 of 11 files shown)


lib/bamboo/template.ex
```diff
@@ -0,0 +1,271 @@
+defmodule Bamboo.Template do
+  @moduledoc """
+  Render emails with layouts, view modules, and templates.
+
+  This module allows rendering emails with layouts and views. Pass an
+  atom (e.g. `:welcome_email`) as the template name to render both HTML and
+  plain text emails. Use a string if you only want to render one type, e.g.
+  `"welcome_email.text"` or `"welcome_email.html"`.
+
+  ## Examples
```
lancejjohnson, line 10: Nitpick: What would you think of renaming this heading to something like "Usage Examples"? My thought is that `## Examples` is already used by Elixir for doctests. When I first saw this I was expecting doctests to follow. Could you change the heading hear to something else for consistency with that convention?

lib/bamboo/template.ex
```diff
@@ -0,0 +1,271 @@
...
+      end
+
+  _Render both text and html emails without layouts._
+
+      defmodule MyApp.Email do
+        use Bamboo.Template, view: MyApp.Email.AccountView
+
+        def welcome_email do
+          new_email()
+          |> render(:welcome)
```
lancejjohnson, line 44:
```suggestion
          render(new_email(), :welcome)
```

Just to conform to community style conventions.

lib/bamboo/template.ex
```diff
@@ -0,0 +1,271 @@
...
+      end
+
+  _Render an email by passing the template string to render._
+
+      defmodule MyApp.Email do
+        use Bamboo.Template, view: MyApp.Email.AccountView
+
+        def html_email do
+          new_email()
+          |> render("html_email.html")
```
lancejjohnson, line 79:
```suggestion
          render(new_email(), "html_email.html")
```

lib/bamboo/template.ex
```diff
@@ -0,0 +1,271 @@
...
+        end
+
+        def text_email do
+          new_email
+          |> render("text_email.text")
```
lancejjohnson, line 84:
```suggestion
          render(new_email(), "text_email.text")
```

lib/bamboo/template.ex
```diff
@@ -0,0 +1,271 @@
...
+          |> put_header("Reply-To", "editors@changelog.com")
+          # This will use the "email.html.eex" file as a layout when rendering html emails.
+          # Plain text emails will not use a layout unless you use `put_text_layout`
+          |> put_html_layout({MyApp.Email.LayoutView, "email.html"})
+        end
+      end
+
+      # lib/my_app/email/views/layout_view.ex
+      defmodule MyApp.Email.LayoutView do
+        use Bamboo.View, path: "/lib/my_app/email/templates/layout"
```
lancejjohnson, line 114: Just to confirm, should these be absolute paths?

lib/bamboo/template.ex
```diff
@@ -0,0 +1,271 @@
...
+
+        def text_email do
+          new_email
+          |> render("text_email.text")
+        end
+      end
+
+  ## HTML Layout Example
+
+      # lib/my_app/email.ex
```
lancejjohnson, line 90: Nitpick: I would expect to find email-related modules is `lib/my_app_web/**/*`. That presumes the Phoenix convention to some extent. But I'm wondering if listing the file paths should either follow the most dominant convention (i.e. Phoenix) or should be left out. Thoughts?

lib/bamboo/template.ex
```diff
@@ -0,0 +1,271 @@
...
+        def different_view_template do
+          new_email
+          # Will use welcome.html template of MyApp.AccountView
+          |> put_view(MyApp.AccountView)
+          |> render("welcome.html")
+        end
+      end
+  """
+  def put_view(email, view) do
+    email |> put_private(:view_module, view)
```
lancejjohnson, line 269:
```suggestion
    put_private(email, :view_module, view)
```

lib/bamboo/view.ex
```diff
@@ -0,0 +1,201 @@
...
+    render_functions = Enum.map(results, &elem(&1, 1))
+
+    quote do
+      unquote(render_functions)
+
+      def render_template(template, _assigns) do
+        raise UndefinedTemplateError,
+          module: __MODULE__,
+          template: template,
+          available: unquote(template_names)
```
lancejjohnson, line 80: Very cool! Like the helpful error messages.

lib/bamboo/view.ex
```diff
@@ -0,0 +1,201 @@
...
+      |> Map.put_new(:html_layout, false)
+      |> Map.put_new(:text_layout, false)
+      |> Map.put_new(:view_module, view_module)
+
+    %{email | private: private}
+  end
+
+  defp merge_assigns(%{assigns: email_assigns} = email, assigns) do
+    assigns = email_assigns |> Map.merge(Enum.into(assigns, %{}))
+    email |> Map.put(:assigns, assigns)
```
lancejjohnson, line 105:
```suggestion
    Map.put(email, :assigns, Map.merge(email_assigns, Enum.into(assigns, %{}))
```

Nitpick: Personally I think this suggestion is more straightforward/readable, but totally your call.

lib/bamboo/view.ex
```diff
@@ -0,0 +1,201 @@
...
+
+    email
+    |> Map.put(:html_body, render_html(email, view_template <> ".html"))
+    |> Map.put(:text_body, render_text(email, view_template <> ".text"))
+  end
+
+  defp render_text_or_html_email(email, template) do
+    cond do
+      String.ends_with?(template, ".html") ->
+        email |> Map.put(:html_body, render_html(email, template))
```
lancejjohnson, line 127:
```suggestion
        Map.put(email, :html_body, render_html(email, template))
```

:)

lib/bamboo/view.ex
```diff
@@ -0,0 +1,201 @@
...
+
+      String.ends_with?(template, ".text") ->
+        email |> Map.put(:text_body, render_text(email, template))
```
lancejjohnson, line 130:
```suggestion
        Map.put(email, :text_body, render_text(email, template))
```

lib/bamboo/view.ex
```diff
@@ -0,0 +1,201 @@
...
+
+      true ->
+        raise ArgumentError, """
```
lancejjohnson, line 133: Is it worth defining a exception here? `InvalidTemplateTypeError`?

lib/bamboo/view.ex
```diff
@@ -0,0 +1,201 @@
...
+    module.render_template(template, email.assigns)
+  end
+
+  defp render_within_layout({layout_view, layout_template}, email, template) do
+    module = email.private.view_module
+    contents = module.render_template(template, email.assigns)
+
+    assigns =
+      email.assigns
+      |> Map.put(:inner_content, contents)
```
lancejjohnson, line 165:
```suggestion
      Map.put(email.assigns, :inner_content, contents)
```

Gotta keep up the fight now that I started it! :)

test/lib/bamboo/phoenix_test.exs
```diff
@@ -1,20 +1,24 @@
 defmodule Bamboo.PhoenixTest do
   use ExUnit.Case
 
-  defmodule LayoutView do
-    use Phoenix.View, root: "test/support/templates", namespace: Bamboo.LayoutView
+  defmodule PhoenixLayoutView do
```
lancejjohnson, line 4: Nice!

test/lib/bamboo/template_test.exs
```diff
@@ -0,0 +1,150 @@
+defmodule Bamboo.TemplateTest do
+  use ExUnit.Case
```
lancejjohnson, line 2: Can this be `async: true`?

lib/bamboo/template.ex
```diff
@@ -0,0 +1,271 @@
...
+      <p>Welcome <%= @person.name %></p>
+
+      # lib/my_app/email/templates/account/welcome_email.text.eex
+      # This will not be rendered within a layout because `put_text_layout` was not used.
+      Welcome <%= @person.name %>
+  """
+
+  import Bamboo.Email, only: [put_private: 3]
+
+  defmacro __using__(view: view_module) do
```
lancejjohnson, line 140: What would you think about accepting layouts as `use` options? To take one of your examples from above:
This:

```elixir
defmodule MyApp.Email do
  use Bamboo.Template, view: MyApp.Email.AccountView
  def welcome_email do
    new_email()
    |> put_text_layout({MyApp.Email.LayoutView, "email.text"})
    |> put_html_layout({MyApp.Email.LayoutView, "email.html"})
    |> render(:welcome) # Pass atom to render html AND plain text templates
  end
end
```

could be used this way:

```elixir
defmodule MyApp.Email do
  use Bamboo.Template,
    view: MyApp.Email.AccountView,
    layout_html: {MyApp.Email.LayoutView, "email.html"},
    layout_text: {MyApp.Email.LayoutView, "email.text"}
  
  def welcome_email do
    # Pass atom to render html AND plain text templates
    render(new_email(), :welcome) 
  end
end
```

I'm thinking that if you define multiple email functions within the module, the layout is set for all of them removing the need to call the layout functions directly each time.

Thoughts?

lancejjohnson (commented):
This looks really good! I think you're adding some really useful changes. Several comments the vast majority of which are nits. Feel free to take or leave.

danbee (approved):

lib/bamboo/template.ex
germsvel: I see what you mean, but I think I'd rather keep it this way. This is very, _very_, similar to our [`Bamboo.Phoenix`](https://hexdocs.pm/bamboo/Bamboo.Phoenix.html#content) docs (copied it from there and then modified it), and Bamboo.Phoenix uses the same `## Examples` language. And I think it's the same pattern we use elsewhere in Bamboo (e.g. [`Bamboo.Adapter` docs](https://hexdocs.pm/bamboo/Bamboo.Adapter.html#content)). So I'm happy leaving it as `## Examples` unless you feel strongly about it.

lib/bamboo/template.ex
germsvel: I actually like the way it is for a few reasons: 

1. Since it's documentation, it's easier for people to read each step separately. The function call within a function call looks less clear at a glance. If it were something like `render(%Email{}, :welcome)` then maybe I'd consider it more. 
2. These functions would typically be used in a longer pipeline, so (I think) there is an _expectation_ of piping to `render/3` in people's minds. I'm not sure I'm explaining this correctly, but there are some functions that are _meant_ to be used with a pipe, regardless of how many arguments have gone before it. They _read_ better when piped into. So our brains are used to thinking of that function as something that gets piped into it (e.g. Wallaby's `assert_has` reads much better piped into `session |> assert_has(text("hello world"))` vs `assert_has(session, text("hello world"))`. Of course, there are functions that are the opposite way (like Phoenix's `link` for example. That would look bad piped into `"hello" |> link(to: Routes....)`)
3. Since this is typically used in pipelines, I think it makes it easier for people to copy into their own code. 

What do you think?

lib/bamboo/template.ex
germsvel: Yeah, curious about your thoughts here. I used the full paths to be explicit. I took inspiration from Phoenix, but in Phoenix you define the `root` (I think you can optionally define a partial `path` too). Then Phoenix uses conventions about your View to determine where to find those templates. 

I like that people can just specify whatever path they want here. That's where those templates are and that's what we compile for that particular view.

lib/bamboo/template.ex
germsvel: I wanted to keep away from any notion of `my_app_web` here because the whole purpose of this feature is for people to use this who _don't_ use Phoenix. Emailing is not always a web concern for people. We use HTML to render some of those emails (though it could also just be text), but not all apps have a "web" layer. What do you think? Still think we should include `my_app_web`?

lib/bamboo/view.ex
germsvel: I think I like the separate steps here. Too much going on in a single line otherwise. I don't mind the temporary `assigns` in line 104. Thanks for the suggestion though. 👍

lib/bamboo/template.ex
germsvel: Yeah, I like this idea. 👍  Let me see what we'd need to do in order to be able to do that but also be able to override the layout. I don't think it'd be a big change.

lib/bamboo/view.ex
germsvel: That's a good question. I think the error is good enough as is. It's an argument error because you passed the wrong argument to `render/2` -> since the template doesn't exist. But I could be convinced otherwise. Not sure I have a strong opinion here.

lib/bamboo/view.ex
germsvel: On second read (now that I'm looking at the full function, I like what you have there. I'll go with it.

lib/bamboo/template.ex
germsvel: I updated this in 84a679435cd47086e85414f8c10fa98ea5d95855 if you'd like to take a look. I like how things turned out.

germsvel:
Someone just pointed me to the fact that `phoenix_view` is being extracted to its own package, and people could use that. That would make it so all of this work is no longer needed. https://github.com/phoenixframework/phoenix/pull/4070

lib/bamboo/template.ex
dotdotdotpaul: The comment doesn't have the leading slash, but the underlying code does.  Maybe be explicit that this is either "always relative to app root", or do something where if there's a leading slash it's absolute (and without the leading slash, it's relative to app root)?

germsvel: Great observation @dotdotdotpaul. I'm curious, is there a better name for what is currently the key `path` that would help make that clear? I'd like to make it clear in documentation _but also_ make it so it's clear from the usage what you're supposed to set there. Have any thoughts?

dotdotdotpaul: Could probably add a config option for "root" and then stipulate that "path" is _always_ relative to whatever "root" is defined as, and if "root" is `nil`, it will use the project root (which is probably the preferred default anyway -- I can only see a different root being used if there's some kind of separate template source).  But I've only had one cup of coffee this morning, so that may not be an entirely baked idea.

germsvel: Re-reading these comments with time, I now realize that what I said at first was totally wrong. We don't take absolute paths. @lancejjohnson I thought you were asking something else, and I got confused. I fixed this in 6297d0db381166c6a0b0744e34e86f966fbe6967. 

@dotdotdotpaul I liked your idea. I was going to implement it, but that's when I realized we don't want absolute paths. So `path` here is just relative paths -- relative to the root of the project. And like you say, that's probably the preferred default anyway.

germsvel:
Even though Phoenix extracted `phoenix_view`, I'm leaning towards including this in Bamboo anyway. It seems like a decent option for people who don't use Phoenix and don't want to bring in `phoenix_view` into their projects. But I'm curious what others think.

Issue #350, opened 2018-02-01
Email preview collapsed html preview


RobinClowers:
When I use the email preview plug, the iframe that should show the html preview gets it's height set to 0, so I don't see the html version. If I change it in dev tools the email is there and rendering correctly.

PR #579, opened 2021-02-15
Collapsed email preview iframe bugfix

almirsarajcic:
As part of the Wallaby feature test I've needed to read an email, but I could only get iframe element using `Wallaby.Query.css("iframe", visible: false)` because the height of the element is 0.

I've tried this in multiple versions of Chrome and chromedriver. Also, I've only been able to experience this in headless Chrome in macOS. I wasn't having this issue with Wallaby option `headless: false` nor on a CI instance running Ubuntu.

I figured this is not important enough that we should have test for this which adds chromedriver dependency, but I can add it if it's required.

Related to https://github.com/thoughtbot/bamboo/issues/350 and https://github.com/thoughtbot/bamboo/pull/351.

### DIFF
--- a/lib/bamboo/plug/sent_email_viewer/index.html.eex
+++ b/lib/bamboo/plug/sent_email_viewer/index.html.eex
@@ -233,7 +233,10 @@
             <script>
             function adjustFrameHeight(iframe) {
               var contentWindow = iframe.contentWindow;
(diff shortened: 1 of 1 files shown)


germsvel (approved):
Thanks for submitting this fix! 🎊

PR #584, opened 2021-02-23
Fix type spec of Mailer.deliver_later/1

lasseebert:
After updating to 2.0.0, my dialyzer fails with a message like this:

```
lib/my_app/mailer/mailgun.ex:17:invalid_contract
The @spec for the function does not match the success typing of the function.

Function:
MyApp.Mailer.Mailgun.BambooMailer.deliver_later/1

Success typing:
@spec deliver_later(_) ::
  {:error,
   <<_::64, _::size(8)>>
   | %{
       :__exception__ => true,
       :__struct__ => Bamboo.EmptyFromAddressError | Bamboo.NilRecipientsError,
       :message => <<_::64, _::size(8)>>
     }}
  | {:ok,
     %{
       :bcc => {binary(), binary()},
       :cc => {binary(), binary()},
       :from => {binary(), binary()},
       :to => {binary(), binary()},
       _ => _
     }}

________________________________________________________________________________
done (warnings were emitted)
```

This PR fixes that spec error.

### DIFF
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -80,7 +80,8 @@ defmodule Bamboo.Mailer do
         Bamboo.Mailer.deliver_now!(config.adapter, email, config, opts)
       end
 
-      @spec deliver_later(Bamboo.Email.t()) :: Bamboo.Email.t()
+      @spec deliver_later(Bamboo.Email.t()) ::
+              {:ok, Bamboo.Email.t()} | {:error, Exception.t() | String.t()}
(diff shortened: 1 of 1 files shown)


lasseebert:
Note that this is not part of https://github.com/thoughtbot/bamboo/pull/583, since it seems that the generated functions inside `__using__` are not dialyzed in that PR.

germsvel (approved):

germsvel:
Good catch! Thanks for opening this PR and thanks for the little note disambiguating from #583. Those kinds of notes are so helpful. 🎉

PR #590, opened 2021-03-10
Fix bug in LocalAdapter with an open_email_in_browser_url config

minhajuddin:
When LocalAdapter is used with an `open_email_in_browser_url` config, delivering emails throws an error https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/adapters/local_adapter.ex#L34-L37 returns a `{"", 0}` which doesn't match https://github.com/thoughtbot/bamboo/blob/master/lib/bamboo/mailer.ex#L206-L209


germsvel:
Thanks! That was fast ⚡

germsvel:
Thanks for this work!

SamuelWillis:
Nice, thanks for this improvement!

PR #593, opened 2021-03-25
Fix example for MandrillHelper.put_param/3

shawnleong:
Thanks for open-sourcing and the work behind bamboo. 
I was perusing the docs and got slightly confused by the example before realising it was an error in the given example.

# Background
The example for Bamboo.MandrillHelper.put_param/3 is incorrectly calling on a non-existent put_param/4
- The example passes email twice to the function
- First through the pipeline
- Second through the function argument

# Fix
This merge request aims to corrects the example in the documentation.


germsvel (approved):
Thank you for submitting this fix! Always very much appreciated.

Issue #594, opened 2021-03-30
Allow overriding of recipients when using in staging environment


davidlibrera:
In one of our staging environment we're working using real addresses and we should avoid sending them emails.
In ruby we usually use gems like [RecipientInterceptor](https://github.com/croaky/recipient_interceptor) that allow to override all recipients (to, cc and bcc)  with a custom list of addresses and setting the original ones in headers.

We have a PR ready that includes a new adapter `RecipientReplacerAdapter` that wraps the real adapter replacing all addresses.
```
config :my_app, MyApp.Mailer,
  adapter: Bamboo.RecipientReplacerAdapter,
  inner_adapter: Bamboo.SendGridAdapter,
  recipient_replacements: ["user-1@example.com", "user-2@example.com],
  ...
```

Could it be an interesting feature?

https://github.com/thoughtbot/bamboo/compare/master...leanpanda-com:master

maartenvanvliet:
A feature to handle use cases such as these is already being worked on 😀

See: https://github.com/thoughtbot/bamboo/pull/591

davidlibrera:
Sorry,  I've looked in Iusses and not in existing PRs

PR #595, opened 2021-03-30
Adapter that replaces all recipients with a default list

davidlibrera:
As I proposed in #594, this PR add a new Adapter that is a wrapper around other adapters (i.e SendgridAdapter).

The idea is that on staging or releases environment, when someone send an email to any addresses, it replaces the to addresses with a default list and sets email headers with the original recipients in order to allow debugging before publishing on the production environment.


davidlibrera:
I close since I can add this behaviour using #591

germsvel:
Thanks @davidlibrera. It's exciting to see that many people are wanting this same feature. 👍

PR #598, opened 2021-04-19
fix deliver_later typo in doc

arghmeleg:
This looks like it's supposed to refer to the non `!` version of this function.

### DIFF
--- a/lib/bamboo/mailer.ex
+++ b/lib/bamboo/mailer.ex
@@ -183,7 +183,7 @@ defmodule Bamboo.Mailer do
   @doc """
   Deliver an email in the background.
 
-  Same as `deliver_later!/2` but does not return an ok tuple and raises on
+  Same as `deliver_later/2` but does not return an ok tuple and raises on
   errors.
 
   If successful, this function only returns the `Email` struct.


lib/bamboo/mailer.ex
```diff
@@ -183,7 +183,7 @@ defmodule Bamboo.Mailer do
   @doc """
   Deliver an email in the background.
 
-  Same as `deliver_later!/2` but does not return an ok tuple and raises on
+  Same as `deliver_later/2` but does not return an ok tuple and raises on
```
germsvel, line 186: I believe this is actually correct as is. 

These docs are for the `deliver_later!/2` function, but here they're saying that the function is the same as `deliver_later/2` (the non-bang version) except `deliver_later!/2` raises on errors. 

Does that make sense? Do you think we could make that more clear?

arghmeleg: @germsvel That's what I was trying to do, remove the `!`, so it reads `Same as deliver_later/2` now.

germsvel: 🤦  I read that backwards, huh? 😅  

Thanks for catching that. Looks good!

JasonMiller:
STEEEEEEVE

PR #600, opened 2021-05-14
Introduce assign/2 to set many assigns

germsvel:
Inspired by https://github.com/thoughtbot/bamboo/pull/563

What changed?
============

We introduce `Bamboo.Template.assign/2` to sets many assigns for an email. It accepts a map or a keyword list as an argument.

We also take inspiration from [Phoenix.LiveView.assign/2][assign/2]

[assign/2]: https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#assign/2

### DIFF
--- a/lib/bamboo/template.ex
+++ b/lib/bamboo/template.ex
@@ -204,7 +204,25 @@ defmodule Bamboo.Template do
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


brimatteng (approved):
Neat!

PR #603, opened 2021-05-21
Use refute_timeout configured for all refute_ assertions

germsvel:
What changed?
============

All refutation test helpers in `Bamboo.Test` should be using the `refute_timeout` configuration when testing on shared mode.

```elixir
config :bamboo, :refute_timeout, 10
```

The `refute_email_delivered_with/1` macro was not using the configuration, but instead it had a hardcoded 100ms. We update it to use the `refute_timeout`, since not using it was a bug.

Finally, the `refute_timeout/0` function currently returns 0ms when Bamboo isn't on shared mode. We change that to be 100ms to match the previously hardcoded value in `refute_email_delivered_with/1` and to match the [default `refute_receive_timeout` from ExUnit](https://hexdocs.pm/ex_unit/ExUnit.html#configure/1-options).

### DIFF
--- a/lib/bamboo/test.ex
+++ b/lib/bamboo/test.ex
@@ -235,16 +235,28 @@ defmodule Bamboo.Test do
   @doc """
   Check that no email was sent with the given parameters
 
-  Similarly to `assert_email_delivered_with`, the assertion waits up to 100ms before
-  failing. Note that you need to send the email again if you want to make other
-  assertions after this, as this will receive the `{:delivered_email, email}` message.
+  Similar to `assert_email_delivered_with/1`, but it checks that an email with
+  those parameters wasn't sent.
+
+  If `Bamboo.Test` is used with shared mode, you must also configure a timeout
+  in your test config.
+
+      # Set this in your config, typically in config/test.exs
+      config :bamboo, :refute_timeout, 10
+
+  The value you set is up to you. Lower values may result in faster tests, but
+  your tests may incorrectly pass if an email is delivered *after* the timeout.
+
+  Note that this assertion helper will grab the email out of the process
+  mailbox. So if you want to make other assertions about the same email after
+  this assertion, you need to send the email again.
 
   ## Examples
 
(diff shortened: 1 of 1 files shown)


lib/bamboo/test.ex
```diff
@@ -255,7 +267,7 @@ defmodule Bamboo.Test do
         receive do
           {:delivered_email, email} -> Map.from_struct(email)
         after
-          100 -> []
+          Bamboo.Test.refute_timeout() -> []
```
brimatteng, line 270: If this is going to be a potentially test breaking concern is it worth a deprecation warning or minor version bump?

germsvel: Yeah, this is a good question. I honestly don't know. Technically, this is a bug that we're fixing, so it could be a patch. But if we consider the existing behavior as correct and us changing this, it could be considered potentially breaking. What do you think?

germsvel: Now that I've thought about this more. Maybe the change we should make is to default the `refute_timeout` to 100ms instead of 0ms. That would make this not a breaking change. Other tests might wait longer if the email fails to deliver, but they pass as soon as the email is received, so 100ms is a better default than 0ms. 

I also like that ExUnit's [refute_receive_timeout](https://hexdocs.pm/ex_unit/ExUnit.html#configure/1-options) defaults to 100ms. So it keeps us consistent with that, which probably matches people's expectations more closely.

brimatteng: That seems like a nice approach to this. They can always change the default explicitly but half a second longer test suite seems like a low price to pay for a suitable default.

brimatteng (approved):

lib/bamboo/test.ex
germsvel: @brimatteng I updated the default to 100ms, so there's no longer a potential for a breaking change, and I updated the PR description. 

I'm also hoping to quickly follow up with a PR that allows users to pass the `timeout` directly, so users should be able to change the timeout per test as desired.

PR #604, opened 2021-05-21
Turn some test helpers from macros -> functions

germsvel:
What changed?
============

We change three test helpers in `Bamboo.Test` from macros into regular functions:

- `assert_delivered_email/1`
- `assert_delivered_email_with/1`
- `refute_email_delivered_with/1`

There are other helpers that are already functions:

- `assert_no_emails_delivered/0`
- `refute_delivered_email/1`

The only macro we leave as a macro (because it needs to be one) is `assert_delivered_email_matches/1`

Why?
----

There seems no reason for those functions to be macros, and they only add extra complexity.

> Macros should only be used as a last resort

- from Elixir's [getting started docs](https://elixir-lang.org/getting-started/meta/macros.html)

A side benefit
-------------

A side benefit of doing this is that we can make all the private functions truly private with `defp`. With the macros, we had to make them public (to access them from the macros), and we set `@doc false` so they wouldn't show up in docs.

### DIFF
--- a/lib/bamboo/test.ex
+++ b/lib/bamboo/test.ex
@@ -463,7 +463,6 @@ defmodule Bamboo.Test do
     !!Application.get_env(:bamboo, :shared_test_process)
   end
 
-  @doc false
   defp normalize_for_testing(email) do
     email
     |> Bamboo.Mailer.normalize_addresses()
--- a/lib/bamboo/test.ex
+++ b/lib/bamboo/test.ex
@@ -157,18 +157,10 @@ defmodule Bamboo.Test do
       unsent_email = Bamboo.Email.new_email(subject: "something else")
       assert_delivered_email(unsent_email) # Will fail
   """
-  defmacro assert_delivered_email(email, opts \\ []) do
-    quote do
-      import ExUnit.Assertions
-      email = Bamboo.Test.normalize_for_testing(unquote(email))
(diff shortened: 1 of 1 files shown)


lib/bamboo/test.ex
```diff
@@ -157,12 +157,9 @@ defmodule Bamboo.Test do
...
   """
-  defmacro assert_delivered_email(email) do
-    quote do
-      import ExUnit.Assertions
-      email = Bamboo.Test.normalize_for_testing(unquote(email))
-      assert_receive({:delivered_email, ^email}, 100, Bamboo.Test.flunk_with_email_list(email))
-    end
+  def assert_delivered_email(email) do
+    email = normalize_for_testing(email)
+    assert_receive({:delivered_email, ^email}, 100, flunk_with_email_list(email))
```
brimatteng, line 162: Do these all need to get updated with the timeout from the other PR?

germsvel: Yes, these will all have to get updated once https://github.com/thoughtbot/bamboo/pull/605 gets merged (or vice versa). I just wanted to introduce them cleanly in separate PRs, and I can resolve the conflicts once I merge one of them.

germsvel: I've rebased this now that the timeout PR has been merged, so this should now be fixed.

brimatteng (approved):
:+1: looks nice and clean

PR #608, opened 2021-06-01
Add inline images in mandrill adapter

agutierrezrodriguez:
This PR adds support to send to mandrill inline images.

### DIFF
--- a/lib/bamboo/adapters/mandrill_adapter.ex
+++ b/lib/bamboo/adapters/mandrill_adapter.ex
@@ -113,7 +113,7 @@ defmodule Bamboo.MandrillAdapter do
     {images, files} =
       attachments
       |> Enum.reverse()
-      |> Enum.split_with(&is_image_inline?/1)
+      |> Enum.split_with(&is_inline_image?/1)
 
     mandrill_message
     |> Map.put(:attachments, format_attachments(files))
@@ -138,10 +138,10 @@ defmodule Bamboo.MandrillAdapter do
     end)
   end
 
-  defp is_image_inline?(%_{content_type: "image/" <> _, content_id: cid}) when not is_nil(cid),
+  defp is_inline_image?(%_{content_type: "image/" <> _, content_id: cid}) when not is_nil(cid),
     do: true
 
-  defp is_image_inline?(_), do: false
+  defp is_inline_image?(_), do: false
 
   defp recipients(email) do
     []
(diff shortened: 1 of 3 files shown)


lib/bamboo/adapters/mandrill_adapter.ex
```diff
@@ -103,12 +103,23 @@ defmodule Bamboo.MandrillAdapter do
...
 
+  defp add_attachments(mandrill_message, %{attachments: attachments}) do
+    {files, images} =
+      attachments
+      |> Enum.reverse()
+      |> Enum.split_with(&is_nil(&1.content_id))
+
+    mandrill_message
+    |> Map.put(:attachments, format_attachments(files))
+    |> Map.put(:images, format_attachments(images))
```
germsvel, line 120: Do you know why we're calling this `images`? Will inline attachments always be images or is that how mandrill categorizes inline attachments?

germsvel (commented):
Thank you for opening this pull request!  

I took a quick glance and left a comment on the changes. So that I can better review the pull request, do you have any links to Mandrill documentation on inline attachments?

lib/bamboo/adapters/mandrill_adapter.ex
agutierrezrodriguez: Yes, of course! https://mailchimp.com/developer/transactional/api/messages/send-new-message/
Inside message attribute at the end.

Is similar to attachments with the difference of send the cid instead of the filename.

germsvel: Thanks for sending that link. That's very helpful! This is the section I'm looking at: 

Screen Shot 2021-06-11 at 8 56 45 AM


Based on what I'm seeing, the changes in this PR seem good. 

My only lingering question is: do you know if it's possible to add non-image files as inline attachments? 

- If that's the case, then someone could add a non-image attachment with `content-id` but we would _incorrectly_ put it in the `:images` field. 
- If Mandrill only allows for images as inline attachments, then I think this PR is good to go.

agutierrezrodriguez: I added a new commit to check the content_type too before add it as inline image.

germsvel (approved):
Thanks so much for all this work! Looks good to me.

agutierrezrodriguez:
Thanks a lot for you @germsvel!! 🍻

Issue #592, opened 2021-03-25
Add support for SendGrid's unique arguments


jason-pather:
SendGrid's unique arguments allow attaching a unique identifier to each email which can be used for tracking:
https://sendgrid.com/docs/for-developers/sending-email/unique-arguments/

SendGrid's Event API emits events that also contain these identifiers, allowing emails and events to be linked:
https://sendgrid.com/docs/for-developers/tracking-events/event/

Support for unique arguments would be a great addition to the SendGridHelper, which already supports a similar feature - adding categories to emails using the `with_categories/2` function:
https://hexdocs.pm/bamboo/Bamboo.SendGridHelper.html#with_categories/2

germsvel:
Thanks for opening this issue @jason-pather. I'd be happy to review a PR for this, if you or others feel up for it. I imagine having a function like `SendGridHelper.with_unique_args(email, args)` would be good.

jeepers3327:
I create a PR to support this one. Please kindly review my PR. 

Thank you.

PR #609, opened 2021-06-01
Add support for Sendgrid unique arguments

jeepers3327:
This PR add support for Sendgrid unique_arguments (#592 )

### DIFF
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -124,13 +124,13 @@ defmodule Bamboo.SendGridAdapter do
     |> put_template_id(email)
     |> put_attachments(email)
     |> put_categories(email)
-    |> put_unique_args(email)
     |> put_send_at(email)
     |> put_settings(config)
     |> put_asm_group_id(email)
     |> put_bypass_list_management(email)
     |> put_google_analytics(email)
     |> put_ip_pool_name(email)
+    |> put_unique_args(email)
   end
 
   defp put_from(body, %Email{from: from}) do
--- a/lib/bamboo/adapters/send_grid_adapter.ex
+++ b/lib/bamboo/adapters/send_grid_adapter.ex
@@ -124,6 +124,7 @@ defmodule Bamboo.SendGridAdapter do
     |> put_template_id(email)
     |> put_attachments(email)
     |> put_categories(email)
+    |> put_unique_args(email)
     |> put_send_at(email)
     |> put_settings(config)
     |> put_asm_group_id(email)
@@ -440,4 +441,12 @@ defmodule Bamboo.SendGridAdapter do
     do: Map.put(body, :ip_pool_name, ip_pool_name)
 
   defp put_ip_pool_name(body, _), do: body
+
+  defp put_unique_args(body, %Email{private: %{unique_args: unique_args}})
+       when is_map(unique_args) do
+    body
+    |> Map.put(:unique_args, unique_args)
+  end
+
+  defp put_unique_args(body, _), do: body
 end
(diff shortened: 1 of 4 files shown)


germsvel (commented):
Hi @jeepers3327 thanks for opening this PR. 

I see that you added the `with_unique_args/2` function to the `SendGridHelper` that puts the args under the private map. That's one side of the equation. 

For this to work, we also need to  grab those unique args and put them in the correct place when sending the email. For example, with categories, we put them in the email's body in https://github.com/thoughtbot/bamboo/blob/6aa74c3ff16604aeba853c34cd525c6b7f46b33f/lib/bamboo/adapters/send_grid_adapter.ex#L126 which calls  https://github.com/thoughtbot/bamboo/blob/6aa74c3ff16604aeba853c34cd525c6b7f46b33f/lib/bamboo/adapters/send_grid_adapter.ex#L305-L311

Do you mind including that in this PR?

jeepers3327:
Apologies, I only thought of the SendgridHelper module.

I've added `with_unique_args` to SendgridAdapter module. Please kindly review.

Thank you.

lib/bamboo/adapters/send_grid_helper.ex
```diff
@@ -294,4 +295,25 @@ defmodule Bamboo.SendGridHelper do
...
+  A map of unique arguments for this email. This will override any existing unique arguments.
+
+  ## Example
+
+      email
+      |> with_unique_args(%{new_arg_1: "new arg 1", new_arg_2: "new arg 2"})
+  """
+  def with_unique_args(email, unique_args) when is_map(unique_args) do
+    unique_args =
+      Map.get(email.private, @unique_args, %{})
```
germsvel, line 309: What do you think about using [`Map.update/4`](https://hexdocs.pm/elixir/1.12/Map.html#update/4) here instead of getting the values and then updating them? I think it reveals intention slightly more clearly.

lib/bamboo/adapters/send_grid_adapter.ex
```diff
@@ -440,4 +441,12 @@ defmodule Bamboo.SendGridAdapter do
     do: Map.put(body, :ip_pool_name, ip_pool_name)
 
   defp put_ip_pool_name(body, _), do: body
+
+  defp put_unique_args(body, %Email{private: %{unique_args: unique_args}})
```
germsvel, line 445: Thanks for adding this function. Could you add a test to ensure this behaves as expected? I know we tested the `with_unique_args` function in the helper, but it would be nice to test this one too.

germsvel (commented):
Looking good. Two really minor comments. I think once those are done we can merge this in.

lib/bamboo/adapters/send_grid_helper.ex
jeepers3327: I tried using `Map.update/4` but I had difficulty on converting my current implementation. 
```
Map.update(email.private, @unique_args, %{}, &Map.merge(&1, unique_args))
```
`Map.update/4` returns not only the values but also the keys and `Email.put_private` expect 3 arguments. Can I ask for your guidance on this one.

germsvel: If you think `Map.update/4` makes it more confusing. I'm okay leaving things as they are. No need to change this. Thanks for looking into it!

lib/bamboo/adapters/send_grid_adapter.ex
germsvel: @jeepers3327 whenever you have time, if we add a test for this, I think the PR is good to go. 👍

jeepers3327: @germsvel I've added a test case for `send_grid_adapter`. Let me know if I need to a add more tests.

germsvel (approved):
Thanks for all the work! 🎉

PR #611, opened 2021-06-14
Improve format of SendGridHelper docs

mrcasals:
This PR iproves the format of the SendGridHelper module docs. 

The first paragraph of the function documentation is used as an excerpt that will be shown in the Summary section of the docs page. For example, before this PR:

image

image

I fixed the first example, and simplified the excerpt for the second. 

The formatting for the `add_personalizations` function was broken due to some missing indentation, see before this PR:

image


germsvel (approved):
🥳 thanks for these changes!

PR #620, opened 2021-08-08
feat: Add Bamboo.TestAdapter.forward/2

jonleighton:
This provides a way to receive emails from other processes without
resorting to shared mode and synchronous tests.

For background, this builds on some work that I did to allow Wallaby
tests to play nicely with Mox, while still allowing the tests to be
async. I wrote about it here:

https://jonleighton.name/2021/asynchronous-browser-tests-with-phoenix/

And then contributed some docs to Wallaby about the setup here:

https://github.com/elixir-wallaby/wallaby/pull/592

This commit basically implements a kind of Mox.allow/3 function for
Bamboo.TestAdapter.

Note that I’ve changed the order of the arguments for this
Bamboo.TestAdapter.forward/2 function. For Mox.allow/3, the owner_pid
comes first:

    Mox.allow(MyMock, owner_pid, child_pid)

But for Bamboo.TestAdapter.forward/2, the child_pid comes first:

    Bamboo.TestAdapter.forward(child_pid, owner_pid)

My reasoning is that in the first example we’re allowing the child_pid
to access mocks defined by the owner pid. But in the second example,
we’re forwarding emails FROM the child_pid TO the owner_pid. So this
order of arguments seemed to make sense to me, but may be slightly
confusing.

### DIFF
--- a/lib/bamboo/adapters/test_adapter.ex
+++ b/lib/bamboo/adapters/test_adapter.ex
@@ -19,6 +19,71 @@ defmodule Bamboo.TestAdapter do
 
   @behaviour Bamboo.Adapter
 
+  use GenServer
+
+  @doc false
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
+  end
+
+  @doc """
+  Forward messages sent in `from_pid` to `to_pid`. This provides a way to
+  write tests that send emails in other processes without resorting to shared
+  mode (which cannot be used with async tests).
+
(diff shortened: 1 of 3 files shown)


germsvel:
Thanks for opening this PR @jonleighton! I really like the idea of what you're doing here. I'll try to find some time later this week to give it a good review.

doomspork:
Hi @jonleighton! Now that BEAM Community is taking over Bamboo we're working through these older PRs. Are you interested in pursuing getting this PR merged? If so, it might be good to review Bamboo and see if any additional changes are necessary now that this PR is 3 years old.

PR #621, opened 2021-08-09
feat: Allow Mailgun base_uri option override

rozanov61:


doomspork:
Hello @rozanov61! Now that BEAM Community has taken over Bamboo I'm working through the older PRs. Could you give me a quick TL;DR on this PR so we can assess if we need to merge this still? It'd be great to understand the context of the change.

Issue #614, opened 2021-06-30
Better local preview for text-only emails


kreba:
Thanks for this gread library!

When viewing an email message in the `Bamboo.LocalAdapter`'s `SentEmailViewerPlug`, an iframe for the HTML preview is always shown and always configured to take up the whole available screen height. 

This is nice in most cases. But for some internal email notifications we are happy with only the text variant. For these messages, the HTML preview is always empty and the real content is only available after scrolling down. This is only a minor nuisance, but still we'd love to see this improved. 

We will happily provide a PR if the maintainers give a green light.

Our idea for an improvement: 

The `.email-detail-bodies-container` in `lib/bamboo/plug/sent_email_viewer/index.html.eex` currently contains the two sections for HTML and text. We suggest to surround both of these sections with a conditional. 
If no HTML content has been added to the message, then reduce the HTML section to a title that says "**No HTML body**" (and omit the iframe altogether). 
And likewise for the text section: If no text body was set, reduce that section to an informative title, too.

germsvel:
Hi @kreba thank you so much for opening this issue! I would be happy to review a PR with those changes. I too have found myself having to scroll down a lot just to find the text version of the email.

PR #622, opened 2021-08-10
Better local preview for text-only emails

kreba:
As promised in #614, this PR makes the local preview more compact for messages that have no HTML body.

Note:
There is a second commit in this PR that adds color to the headings "HTML body" and "Text body". This is not exactly related to the issue at hand and we understand that this change might not be welcome. Feel free to leave it out.


silvagustin:
Unfortunatelly, this is not working for [Bamboo Phoenix](https://github.com/thoughtbot/bamboo_phoenix) and I think the problem is that the `Bamboo` dependency version [used in there](https://github.com/thoughtbot/bamboo_phoenix/blob/main/mix.exs) is an old one which doesn't have this change.

Issue #517, opened 2020-01-27
Serve SentEmailViewer static assets from a plug instead of CDN


seivan:
Seems like [`normalize.css`](https://github.com/thoughtbot/bamboo/tree/master/lib/bamboo/plug/sent_email_viewer) is served from some CDN. I'm a bit uncomfortable with having assets that we don't control, it would be nice if there was an opt in feature to allow us to serve that ourselves from a custom plug

Maybe there could be a config like: 
```elixir
config :my_app, MyApp.Mailer,
  adapter: Bamboo.LocalAdapter
  assets: "static"
```

Then you could just serve the file from `/priv/static` with a plug: 
```elixir
plug Plug.Static,
  at: "/assets/", from: :api, gzip: true,
  only: ~w(css fonts images js favicon.ico robots.txtn)
```

I don't mind wiring up a PR for this, if it's acceptable.

germsvel:
Hi @seivan, I think this optional configuration could make sense and we can fallback to the CDN. It would be nice for it to work out of the box without configuration. Of course if you don't add normalize, then the preview emails might look different or slightly broken. I'm not actually sure to what extent the preview styles rely on normalize.css.

PR #625, opened 2021-08-27
Remove CDN & Implement in-use css from CDN into internal css

DhananjayPurohit:
Created this pull request in response to issue #517. Removed CDN from all the files serving UI for SentEmailViewer and implemented css being in use from CDN into internal css.


DhananjayPurohit:
Thanks @germsvel for giving it a review!

Attaching the screenshots taken while running on Chrome along with the inspect window.

Before making changes:-

Screenshot (298)
Screenshot (299)

 After making changes:-

Screenshot (300)
Screenshot (301)

As you can see the inspect window having CDN before making any changes and replacing the CDN on making changes. I would love to test on other browsers too if there is any possibility of breaking on any browser.

Issue #430, opened 2018-09-05
Any options for inline attachments MailgunAdopter


ijunaid8989:
In past, we were using mailgun wrapper https://github.com/evercam/mailgun/commit/8c73ada08137ce875b487949167d60fecb1715dc and added inline attachment options by ourselves.

Now we are planning to switch to Bamboo, and I have done formal emails method conversion from mailgun to bamboo mailgun adopter but it's not sending inline attachments, Is there any option for that in bamboo? 

For emails like 

![image](https://user-images.githubusercontent.com/13561999/45094751-a423e900-b135-11e8-84e3-e862abcf7bfe.png)

Instead of showing image as mentioned in template.

```html
<img src="cid:snapshot.jpg" alt='Last Snapshot' style="width: 100%; display:block; margin:0 auto;">
```

its just showing it alt text. and in the bottom it attached the snapshot. I am sorry if that is not the right place to ask this. I have tried in Elixir Slack channel but no one is interested to answer.

germsvel:
Hi @ijunaid8989 thanks so much for opening the issue and for the patience! I'm just starting to maintain Bamboo and trying to look through the open issues. 

This seems related to https://github.com/thoughtbot/bamboo/issues/502. It seems like it might be good to allow people to set the `content_disposition` and then map that to the adapters as needed. 

If this is still something of interest, let me know. I'll try to get to it, or if you're up for it, I'd be happy to review a PR.

PR #631, opened 2021-10-22
feat: Add arbitrary header overrides for attachments

### LINKED ISSUE #502: Add attachment `content-disposition` and `content-id` fields
SendGrid api has the following properties for emails:

https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/index.html

See the `disposition` and `content_id` properties under attachments.

These turn out to be pretty important when you're sending .ics files (calendar invites). By default SendGrid populates the disposition value with "attachment" if it isempty, but Microsoft Outlook requires it to be set to "inline" in order to display the .ics file as a "native" invitation.

I think since there's a content_type field in the `%Bamboo.Attachment{}` struct already, it makes sense to also be able to set a `content_disposition` field.

(Note that when setting disposition to "inline", Sendgrid also requires a unique `content_id` to be set.)

jarrodmoldrich:
This allows bamboo adapters to be accept overriding headers for special
use cases. **Updates to those adapters would be required to use this
functionality.**

For e.g., in the specific use-case of adding inlined attachments, some
special case headers are required. Now it's possible to add
`X-Attachment-Id` and override the Content-Disposition header with
`inline` to permit the proper display of inlined content, without an
email client showing them as explicit attachments for the user to
download.

Closes #502 #430 

Related PRs:
https://github.com/kalys/bamboo_ses/pull/43
https://github.com/DockYard/elixir-mail/pull/137


doomspork:
Hi @jarrodmoldrich! Now that BEAM Community is taking over Bamboo we're working through these older PRs. Would you like to update this PR or should we close?

jarrodmoldrich:
@doomspork Hi there!  I probably won't be testing this against SES as I've retired the infra I was using, but I'll get the PR up-to-date and to a working state in the coming weeks.

doomspork:
Amazing, thank you @jarrodmoldrich!

jarrodmoldrich:
@doomspork This is ready for another look (assuming it passes your CI)

jarrodmoldrich:
@doomspork Has this work been deprecated?  I did update the PR but received no feedback on it.

PR #639, opened 2022-06-28
Display attached images if referenced in html

alappe:
Looks for cid:… references in the html_body and replaces them with the local attachment url, if a match is found.


doomspork:
Hi @alappe! Are you still interested in pursuing these changes? If so we'd like to add tests to all new changes when applicable. 🙏

PR #655, opened 2023-03-31
feat: SendGrid `subscription_tracking` setting

warnero:
* Adding in missing `subscription_tracking` setting

This adds the capability to send a value for `subscription_tracking` that is included in the `tracking_settings` section of the SendGrid mail body


doomspork:
@warnero I'll take a look at this, and remove the stale tag, now that the BEAM Community team has taking over Bamboo 🫡

PR #663, opened 2024-05-03
Print email metadata on preview without body

pfac:
Why:

* In a setup where one uses a email service with transactional templates or any other mechanism where the system does not actually render the email content but instead triggers an external service to generate the emails, previewing sent emails in a development environment is pretty much useless since nothing is shown. Instead developers have to use the logs to try and parse the relevant email information.

This change addresses the issue by:

* Including the email information as metadata when the email has no body. This shows the data being sent to external service to generate the email, thus providing a way for developers to check the relevant values in a development environment.

Screenshot 2024-05-04 at 00 02 12

### DIFF
--- a/lib/bamboo/plug/sent_email_viewer/index.html.eex
+++ b/lib/bamboo/plug/sent_email_viewer/index.html.eex
@@ -257,6 +257,11 @@
           <% else %>
             <h3 class="email-detail-body-label">No text body</h3>
           <% end %>
+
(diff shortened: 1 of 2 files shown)


pfac:
Just noticed this was closed by accident when we merged the change locally.

pfac:
Ping, is anyone looking at these?

yordis (approved):
Since this is only used for development purpose; I am OK with it.

btkostner (approved):
Seems reasonable

PR #668, opened 2024-06-27
refactor: modernization

doomspork:


.formatter.exs
```diff
@@ -1,7 +1,7 @@
 # This file is synced with beam-community/common-config. Any changes will be overwritten.
 
 [
-  import_deps: [],
+  import_deps: [:plug],
```
doomspork, line 4: Need to see why common-config didn't detect plug.

btkostner (approved):
Mostly formatting changes. No code related changes. I looked over all of the tests to ensure it's only formatting and the tests still pass so we should be good to go.

PR #688, opened 2024-11-26
Add support for content_id in SendgridAdapter

hubertlepicki:
Introduces support for content_id so that inline images in e-mails can be send and referenced from HTML content of the e-mail, when using SendGrid Adapter.


hubertlepicki:
knock knock

doomspork (commented):
Hi @hubertlepicki! Apologizes for the delay. 

I'm trying to figure out why the GHAs won't run on your branch before I merge. I'd like to make sure all the checks are passing and I can tell they're broken because we're using conventional commits now and your PR should fail that check.

Once I get GHA sorted and we get things ✅ we should be good to go

doomspork:
Merged this commit via #691. Not sure why these GHAs aren't queuing.

PR #769, opened 2026-08-14
Replace Hackney with Req 

pxp9:
This PR replaces Hackney with Req, it passes all the testing previous suite.

## Motivation

Req is the standard the facto in elixir ecosystem for http calls, it is actively maintained by Dashbit and it does not suffer performance issues under big load.


yordis (changes requested):
I agree with the intent, but this would require a breaking change, since you are changing the default.

Also, we would still provide the `:hackney` for those using it, intentionally. So at least, we need to provide that. Coming from some scars dealing with multi-adapters, they do not behave the same, and I don't know what I don't know.

pxp9:
hey @yordis , if i refactor the library to allow adapters and i build multiple adapters , would you approve this change?

the idea is make configuration and make hackney and req optional deps so you can swap the adapters easily.

pxp9:
the idea is to standarize in a module all the http requests and make 2 adapters to make it behave the same.

yordis:
I will defer to @doomspork to make a decision here.

This package is quite popular across the ecosystem and stable enough. I am not sure if I want to be responsible for the breaking change without a real proven advantage other than being aligned (which I understand your intent).

Req comes with all the configuration built-in, so any default change to the configuration, any extra or removed middleware means a potential breaking change for this package. I would lean more in Finch since all the middleware/step and extra configuration is added complexity (Req is built on top of Finch).

Lastly,

Some discussion exists around bringing `:gen_http` to OTP itself, I would lean more towards such a package in the future, since this isn't something that takes advantage of something like Req; we want a stable, slow-moving machine here.

pxp9:
I was thinking we can do multiple adapters

```
http.ex
|--- hackney.ex
|--- finch.ex
| --- req.ex
```

and refactor the rest of the code to work with http.ex and some options
make an option called adapter and refactor the option for passing the options to hackney to pass to the specific adapter, so that option will work the same as it is working right now.

and the default adapter could be hackney so there is no breaking change there.

if we assume hackney defaults, the library will still work as expected for most users, but we allow users to progressively change to other adapters or new users to use the adapter they want.

this is similar to what most elixir ecosystem is doing, take ecto as an example, ecto has multiple adapters for each database and it is not so hard to maintain, because it does not require a lot of time to Jose to keep ecto working for all DBs.

each of the adapters could have an option dependency so we allow the user to only pull the adapter it needs.

this would be the only breaking, but as long as we document this in the release, i do not see any issue, the work of bamboo user when he updates is to read the release notes and test the application when updating bamboo.

What do you think @yordis ?

would you approve that ?

yordis:
@pxp9 I would prefer to start with the Problem-Statement from runnable production software. Every new abstraction means something new I am personally responsible for.

yordis:
@doomspork feel free to overule me, but I am not taking this particular PR. I would rather prefer to move the discussions into Issues first.

